PR #26, opened 2019-10-29
Issue/16 resource serialization

### LINKED ISSUE #16: Serialize and deserialize resources
prior to deploy, to get the same semantics as on the server

https://github.com/inmanta/openstack/blob/7bcf5bf0ecb6ad9dabdffe1a0370ff3e527b34e7/tests/common.py#L9

andraskvr:
Closes #16

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -277,6 +277,7 @@ class Project():
             if not apply_filter(resource):
                 continue
 
+            resource = self.check_serialization(resource)
             return resource
 
         return None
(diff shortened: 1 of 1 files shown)


pytest_inmanta/plugin.py
```diff
@@ -300,6 +302,7 @@ def dryrun(self, resource, run_as_root=False):
     def deploy_resource(self, resource_type: str, status=const.ResourceState.deployed, run_as_root=False, **filter_args: dict):
         res = self.get_resource(resource_type, **filter_args)
         assert res is not None, "No resource found of given type and filter args"
+        res = self.check_serialization(res)
```
wouterdb: why not add it to get_resource directly, such that test that use the resource, but not the standard deploy also get the correct resource?

andraskvr: Sure thing

PR #39, opened 2020-02-13
Added --install_mode option

wouterdb:
did not add testcases, as this would be quite complex compared to the functionality

### DIFF
--- a/README.md
+++ b/README.md
@@ -105,10 +105,11 @@ A test case, to test this plugin looks like this:
 
 The following options are available.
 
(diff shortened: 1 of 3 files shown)


arnaudsjs:
Open question: If an incorrect value is provided to the `install_mode` option, due to a typo. It will result in a warning but not a hard failure. Is this desired behavior for pytest-inmanta? Do we need to change this into a hard failure?

wouterdb:
I have asked myself the same question. 

@bartv wdyt?

bartv:
I prefer a hard fail as well

wouterdb:
@arnaudsjs I added a hard failure on bad options

PR #41, opened 2020-02-20
Return handler logs

### LINKED ISSUE #35: Explain how to get logs from a deployment
add documentation and/or helper methods to get logs from a handler.

andraskvr:
Closes #35

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -320,7 +320,12 @@ class Project():
     def dryrun(self, resource, run_as_root=False):
         return self.deploy(resource, True, run_as_root)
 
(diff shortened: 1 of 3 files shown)


pytest_inmanta/plugin.py
```diff
@@ -336,15 +341,22 @@ def deploy_resource(self, resource_type: str, status=const.ResourceState.deploye
 
         assert ctx.status == status
         self.finalize_context(ctx)
+        if return_logs:
+            return res, ctx.logs
```
wouterdb: I would prefer not to have variable number of return arguments. 

It would be cleaner to keep the context in a variable and have a get_last_logs and get_last_context 

@bart wdyt?

wouterdb: Alternative would be to have another method name for the one that exposes its context.

PR #44, opened 2020-03-23
Support testing of regular functions and classes in modules

### LINKED ISSUE #37: Add feature to allow testing regular functions, classes
Right now there is no possibility to test regular functions and classes from the level of a project.

It would be very useful to add a feature that would enable testing any item in the plugins/__init__.py

This should be done by either
1. a normal import
2. a method on the project fixture which return the module 

e.g.

```python
def my_function(arg):
    pass

class MyClass:
    pass
```
```python
mymodule = project.do_import('mymodule')

mymodule.my_function
```

andraskvr:
The importing mechanism was implemented in inmanta/inmanta#507, this PR illustrates it with a test case.
Closes #37

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,15 @@ And dryrun
     assert changes == {"value": {'current': 'read', 'desired': 'write'}}
 ```
 
+Testing functions and classes defined in a module is also possible 
+by simply importing them after the project is compiled
+
+```python
+project.compile("") # Model to compile
+from inmanta_plugins.testmodule import regular_function
+
+regular_function("example")
+```
 ## Testing plugins
 
 Take the following plugin as an example:
--- a/README.md
+++ b/README.md
@@ -67,13 +67,13 @@ And dryrun
 ```
 
 Testing functions and classes defined in a module is also possible 
-by simply importing them after the project is compiled
+by simply importing them after the project is loaded or compiled
 
 ```python
-project.compile("") # Model to compile
-from inmanta_plugins.testmodule import regular_function
-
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -66,6 +66,15 @@ And dryrun
     assert changes == {"value": {'current': 'read', 'desired': 'write'}}
 ```
 
+Testing functions and classes defined in a module is also possible 
+by simply importing them after the project is compiled
+
+```python
+project.compile("") # Model to compile
```
arnaudsjs: Compiling an empty model looks quite strange and can be quite confusing. In my opinion the functions and classes should be loaded automatically, as such that the compile step is not required anymore. @wouterdb wdyt?

wouterdb: It is a bit of a tricky situation.

1. The module loader that is now in the core allows loading plugins from modules that are not actually imported (but just happen to be there on disk) (this is why this works without an import in the main file)
2. By simply enabling the module loader, the import will just work, unless there are dependencies on other modules that are required by this one, but that haven't been downloaded.
3. The get_plugin_function call could now also get this problem, that plugins can fail to load due to dependencies on other modules. 

For this case, I would opt for the solution that @andraskvr has implemented now, but I would not use an empty main file, but one that imports the module under test. 

For the get_plugin_function, to work on the same principle would be a huge performance regression. 

So there is a case to be made to do the load when the global fixture is set up, from that point on, the modules can always be loaded.  It will change the way pytest fails when the project fails to load, but not too much.

PR #48, opened 2020-04-01
Close handler caches properly

### LINKED ISSUE #42: Properly close caches on handlers
Currently the handler cache is not closed.

- make a finalize_handler method that properly dispose of a handler, for use in testcases
- ensure that the cleanup method clears all caches.

andraskvr:
Closes #42

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -20,7 +20,6 @@ import tempfile
 import os
 import shutil
 import sys
-import io
 import logging
 import imp
 import glob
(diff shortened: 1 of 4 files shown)


pytest_inmanta/plugin.py
```diff
@@ -536,4 +537,16 @@ def check_serialization(self, resource: Resource) -> Resource:
 
     def clean(self) -> None:
         shutil.rmtree(os.path.join(self._test_project_dir, "libs", "unittest"))
+        self.finalize_all_handlers()
         self.create_module("unittest", initcf=get_module_data("init.cf"), initpy=get_module_data("init.py"))
+
+    def finalize_handler(self, handler: ResourceHandler) -> None:
```
wouterdb, line 543: The deploy() method should also call this method.

pytest_inmanta/plugin.py
```diff
@@ -536,4 +537,16 @@ def check_serialization(self, resource: Resource) -> Resource:
...
+        versions = sorted(handler.cache.counterforVersion.keys())
+        for version in versions:
+            if handler.cache.is_open(version):
```
wouterdb, line 546:
```suggestion
            while handler.cache.is_open(version):
```

this has reference counting and may need to be closed multiple times.

PR #55, opened 2020-05-25
fixed import error for module that imports other modules from its plugins

sanderr:


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# V 1.2.1
```
arnaudsjs, line 1: Why v1.2.1? The latest stable release is v1.1.0.

sanderr: I think I've gotten too used to pull requests on modules.

PR #63, opened 2020-06-24
Add meaningful error message when --venv doesn't exist

### LINKED ISSUE #62: Show clear error when --venv points to a non-existing directory
Currently, the following error is shown when --venv points to a non-existing directory:

```
Traceback (most recent call last):
  File "/venv/lib64/python3.6/site-packages/inmanta/env.py", line 84, in init_env
    venv.create(path, clear=True, with_pip=True)
  File "/usr/lib64/python3.6/venv/__init__.py", line 344, in create
    builder.create(env_dir)
  File "/usr/lib64/python3.6/venv/__init__.py", line 59, in create
    context = self.ensure_directories(env_dir)
  File "/usr/lib64/python3.6/venv/__init__.py", line 106, in ensure_directories
    create_if_needed(env_dir)
  File "/usr/lib64/python3.6/venv/__init__.py", line 95, in create_if_needed
    os.makedirs(d)
  File "/usr/lib64/python3.6/os.py", line 220, in makedirs
    mkdir(name, mode)
FileExistsError: [Errno 17] File exists: '/tmp/tmp6nc1u120/.env'
```

andraskvr:
Closes #62

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,9 +1,6 @@
-# V 1.5.0
-Changes in this release:
-- Added meaningful error message when --venv points to a non-existing directory (#62)
-
 # V 1.4.0
 Changes in this release:
+- Added meaningful error message when --venv points to a non-existing directory (#62)
 
 # V 1.3.0
 Changes in this release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# V 1.5.0
+Changes in this release:
+- Added meaningful error message when --venv points to a non-existing directory (#62)
+
 # V 1.4.0
 Changes in this release:
 
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -176,6 +176,8 @@ def project_factory(request):
         modulepath.append(str(Path(os.getcwd()).parent))
 
     env_override = get_opt_or_env_or(request.config, "inm_venv", None)
+    if env_override and not os.path.exists(env_override):
+        raise Exception("Specified venv %s does not exist", env_override)
     if env_override is not None:
         try:
             os.symlink(env_override, os.path.join(test_project_dir, ".env"))
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -176,7 +176,7 @@ def project_factory(request):
         modulepath.append(str(Path(os.getcwd()).parent))
 
     env_override = get_opt_or_env_or(request.config, "inm_venv", None)
-    if env_override and not os.path.exists(env_override):
+    if env_override and not os.path.isdir(env_override):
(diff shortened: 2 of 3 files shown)


pytest_inmanta/plugin.py
```diff
@@ -176,6 +176,8 @@ def project_factory(request):
         modulepath.append(str(Path(os.getcwd()).parent))
 
     env_override = get_opt_or_env_or(request.config, "inm_venv", None)
+    if env_override and not os.path.exists(env_override):
```
arnaudsjs, line 179: `env_override` should be a directory, so I would use `os.path.isdir` instead.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# V 1.5.0
```
arnaudsjs, line 1: This should be in the v1.4.0 section I guess?

tests/test_options.py
```diff
@@ -26,3 +28,11 @@ def test_module_in_place(testdir):
...
 
     assert os.path.exists(os.path.join(path, "testfile"))
+
+
+def test_not_existing_venv_option(testdir):
+    testdir.copy_example("testmodule")
+
+    result = testdir.runpytest("tests/test_resource_run.py", "--venv", os.path.join(tempfile.gettempdir(), str(uuid.uuid4())))
+
+    result.assert_outcomes(error=1)
```
arnaudsjs, line 38: With or without this patch, this test case will succeed. The goal of this ticket is to improve the error reporting, so it would be better to create an assertion on the exception message.

tests/test_options.py
```diff
@@ -26,3 +28,11 @@ def test_module_in_place(testdir):
     result.assert_outcomes(passed=1)
 
     assert os.path.exists(os.path.join(path, "testfile"))
+
+
+def test_not_existing_venv_option(testdir):
+    testdir.copy_example("testmodule")
+
+    result = testdir.runpytest("tests/test_resource_run.py", "--venv", os.path.join(tempfile.gettempdir(), str(uuid.uuid4())))
```
arnaudsjs, line 36: It's better to use the `tmpdir` fixture. This will give you an empty temporary directory. `os.path.join(tmpdir, str(uuid.uuid4()))` is guaranteed to not exist, while os.path.join(tempfile.gettempdir(), str(uuid.uuid4())) only is very likely to not exist.

pytest_inmanta/plugin.py
```diff
@@ -176,6 +176,8 @@ def project_factory(request):
         modulepath.append(str(Path(os.getcwd()).parent))
 
     env_override = get_opt_or_env_or(request.config, "inm_venv", None)
+    if env_override and not os.path.isdir(env_override):
+        raise Exception("Specified venv %s does not exist", env_override)
```
arnaudsjs, line 180:
```suggestion
        raise Exception(f"Specified venv {env_override} does not exist")
```

PR #91, opened 2020-08-25
Issue/66 project no plugins deprecate

### LINKED ISSUE #66: The project_no_plugins fixture does load plugins
an alternative has been implemented using env variables, this fixture can be deprecated

sanderr:
# Description

Deprecated `project_no_plugins` fixture

closes #66

# Self Check:

- [x] Changelog entry
- [x] ~~End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )~~

# Reviewer Checklist:

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -23,6 +23,7 @@ import os
 import shutil
 import sys
 import tempfile
+import warnings
 from collections import defaultdict
 from distutils import dir_util
 from pathlib import Path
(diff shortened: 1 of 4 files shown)


wouterdb (approved):

pytest_inmanta/plugin.py
```diff
@@ -138,6 +138,10 @@ def project(project_shared, capsys):
 
 @pytest.fixture()
 def project_no_plugins(project_shared_no_plugins, capsys):
+    LOGGER.warning(
+        "The project_no_plugins fixture is deprecated in favor of the %s environment variable.",
+        option_to_env["inm_no_load_plugins"],
+    )
```
arnaudsjs, line 144: This is just a normal log message and not a deprecation warning? If I am not mistaken, deprecationwarning should be logged via the `warnings.warn()` method. Probably your test case is catching another deprecationwarning than this one.

arnaudsjs (changes requested):

pytest_inmanta/plugin.py
sanderr: You're right. I started out wanting to raise a `DeprecationWarning` but then noticed the logger and decided to use that instead. I actually have little experience here? Should all warnings use the logger or should I raise an actual `DeprecationWarning`?

arnaudsjs: It seems that the deprecationwarning is ignored by python by default, so to me it looks better to use the standard logger.

sanderr: I think pytest explicitly warns about deprecation warnings, which is what counts here. I've just discovered that the logger's warning is actually ignored by pytest, so I guess I'll go with the `DeprecationWarning` then.

arnaudsjs (approved):

PR #72, opened 2020-08-03
Issue/46 get plugin improved

sanderr:
# Description

- Improves `Project._load_plugins()` implementation to make sure it is compatible with the compiler's import mechanism
- Adds `inmanta_plugins` fixture to make abstraction of required module reloading when the compiler project is reset

closes #46 
closes #49

# Self Check:

Strike through any lines that are not applicable (`~~line~~`) then check the box

- [x] Attached issue to pull request
- [x] Changelog entry
- [x] Code is clear and sufficiently documented
- [x] Sufficient test cases (reproduces the bug/tests the requested feature)
- [x] Correct, in line with design
- [x] End user documentation is included or an issue is created for end-user documentation

# Reviewer Checklist:

- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)
- [ ] Code is clear and sufficiently documented
- [ ] Correct, in line with design

PR #115, opened 2020-10-01
Improve documentation of options

### LINKED ISSUE #67: Improve documentation on all available configuration options
This ticket should:

- Ensure that each configuration option and corresponding environment variable is clearly explain in the documentation.
- For the less intuitive options (e.g. --venv, --use-module-in-place, --no_load_plugins, etc.), the documentation should also explain why that option was introduced. In other words, which problem(s) does that option solve.

andraskvr:
# Description

closes #67 

# Self Check:

- [x] Changelog entry
- [ ] ~Code is clear and sufficiently documented~
- [ ] ~Sufficient test cases (reproduces the bug/tests the requested feature)~

# Reviewer Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,8 +140,9 @@ The following options are available.
     Using this option can speed up the tests, because the module dependencies are not downloaded multiple times.
  * `--module_repo`: location to download modules from, overrides `INMANTA_MODULE_REPO`. The default value is the inmanta github organisation.
  * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`.
- * `--no-load-plugins`: Don't load plugins in the Project class. Overrides `INMANTA_TEST_NO_LOAD_PLUGINS`. 
- When not using this option during the testing of plugins, it's possible that the module's `plugin/__init__.py` is loaded multiple times, 
+ * `--no_load-plugins`: Don't load plugins in the Project class. Overrides `INMANTA_TEST_NO_LOAD_PLUGINS`. 
+ When not using this option during the testing of plugins with the `project.get_plugin_function` method, 
+ it's possible that the module's `plugin/__init__.py` is loaded multiple times, 
  which can cause issues when it has side effects, as they are executed multiple times as well.
  
  Use the generic pytest options `--log-cli-level` to show Inmanta logger to see any setup or cleanup warnings. For example,
--- a/README.md
+++ b/README.md
@@ -134,11 +134,15 @@ The following options are available.
 
  * `--venv`: folder in which to place the virtual env for tests (will be shared by all tests), overrides `INMANTA_TEST_ENV`.
    This options depends on symlink support. This does not work on all windows versions. On windows 10 you need to run pytest in an
-   admin shell.
+   admin shell. Using a fixed virtual environment can speed up running the tests.
  * `--use-module-in-place`: makes inmanta add the parent directory of your module directory to it's directory path, instead of copying your
-    module to a temporary libs directory.
+    module to a temporary libs directory. It allows testing the current module against specific versions of dependent modules. 
+    Using this option can speed up the tests, because the module dependencies are not downloaded multiple times.
  * `--module_repo`: location to download modules from, overrides `INMANTA_MODULE_REPO`. The default value is the inmanta github organisation.
- * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`  
+ * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`.
+ * `--no-load-plugins`: Don't load plugins in the Project class. Overrides `INMANTA_TEST_NO_LOAD_PLUGINS`. 
+ When not using this option during the testing of plugins, it's possible that the module's `plugin/__init__.py` is loaded multiple times, 
+ which can cause issues when it has side effects, as they are executed multiple times as well.
  
  Use the generic pytest options `--log-cli-level` to show Inmanta logger to see any setup or cleanup warnings. For example,
  `--log-cli-level=INFO`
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -134,11 +134,15 @@ The following options are available.
...
  * `--use-module-in-place`: makes inmanta add the parent directory of your module directory to it's directory path, instead of copying your
-    module to a temporary libs directory.
+    module to a temporary libs directory. It allows testing the current module against specific versions of dependent modules. 
+    Using this option can speed up the tests, because the module dependencies are not downloaded multiple times.
  * `--module_repo`: location to download modules from, overrides `INMANTA_MODULE_REPO`. The default value is the inmanta github organisation.
- * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`  
+ * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`.
+ * `--no-load-plugins`: Don't load plugins in the Project class. Overrides `INMANTA_TEST_NO_LOAD_PLUGINS`. 
+ When not using this option during the testing of plugins, it's possible that the module's `plugin/__init__.py` is loaded multiple times, 
+ which can cause issues when it has side effects, as they are executed multiple times as well.
```
sanderr, line 145: This is not (should not) be accurate anymore with the latest pytest-inmanta. At this point I actually can't think of a reason you would run with this option enabled, unless it speeds up test runs.
Also, I think it should be clarified what "plugin loading" means, because to someone unfamiliar with this term in the context of pytest-inmanta it might seem to refer to the plugin loading by the compiler. Plugin loading in this case only refers to the loading of plugins by pytest-inmanta for use with `project.get_plugin_function(function_name: str)`. It might even be fair to say that this method is no longer the recommended approach for testing plugins since the `inmanta_plugins` package and fixture with latest core and latest pytest-inmanta.
Concluding, I think the description for this option should at least mention the `project.get_plugin_function`.

sanderr (changes requested):
These clarifications should also be added to the option documentation in `plugin.py` so they are returned by `pytest --help`.

README.md
andraskvr: Do you think the `--no-load-plugins` option should be deprecated instead? 
Because in that case, detailing the problem that's already solved in another way might not be necessary at all.

sanderr: That's a good point. From what I know of it I think it might indeed be deprecated, but it's been around from before the issues it caused for me so there might be motivations I don't know about. I think we should pull in @arnaudsjs here.

arnaudsjs: The option `--no-load-plugins` was introduced as a hotfix (https://github.com/inmanta/pytest-inmanta/pull/60). If I am not mistaken, the root cause of the issue was resolved in https://github.com/inmanta/pytest-inmanta/pull/72. So if this assumption is true, the option can be deprecated. But let's first verify that the problem has been resolved by removing this option from all our internal projects to see what happens. @andraskvr Do you create a ticket for that?

README.md
```diff
@@ -134,11 +134,15 @@ The following options are available.
...
-   admin shell.
+   admin shell. Using a fixed virtual environment can speed up running the tests.
  * `--use-module-in-place`: makes inmanta add the parent directory of your module directory to it's directory path, instead of copying your
-    module to a temporary libs directory.
+    module to a temporary libs directory. It allows testing the current module against specific versions of dependent modules. 
+    Using this option can speed up the tests, because the module dependencies are not downloaded multiple times.
  * `--module_repo`: location to download modules from, overrides `INMANTA_MODULE_REPO`. The default value is the inmanta github organisation.
- * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`  
+ * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`.
+ * `--no-load-plugins`: Don't load plugins in the Project class. Overrides `INMANTA_TEST_NO_LOAD_PLUGINS`.
```
arnaudsjs, line 143:
```suggestion
 * `--no_load_plugins`: Don't load plugins in the Project class. Overrides `INMANTA_TEST_NO_LOAD_PLUGINS`. 
```

README.md
sanderr: I think this will require a release of pytest-inmanta.

arnaudsjs (changes requested):

README.md
andraskvr: I created #116 for this, please extend the description, if I left something out.

sanderr (approved):

README.md
```diff
@@ -134,11 +134,16 @@ The following options are available.
...
-   admin shell.
+   admin shell. Using a fixed virtual environment can speed up running the tests.
  * `--use-module-in-place`: makes inmanta add the parent directory of your module directory to it's directory path, instead of copying your
-    module to a temporary libs directory.
+    module to a temporary libs directory. It allows testing the current module against specific versions of dependent modules. 
+    Using this option can speed up the tests, because the module dependencies are not downloaded multiple times.
  * `--module_repo`: location to download modules from, overrides `INMANTA_MODULE_REPO`. The default value is the inmanta github organisation.
- * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`  
+ * `--install_mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`.
+ * `--no_load-plugins`: Don't load plugins in the Project class. Overrides `INMANTA_TEST_NO_LOAD_PLUGINS`.
```
arnaudsjs, line 143: Also mention that the value of the `INMANTA_TEST_NO_LOAD_PLUGINS` environment variable has to be an non-empty string to not load the plugins.

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #200, opened 2021-06-07
project.compile logs the model with line numbers

### LINKED ISSUE #199: Dedent input model from compile call and print it with line numbers on compile failure
This has been done in before in systemtenant:
```python
        model = dedent(model.strip("\n"))
        line_numbers_model = self._enumerate_model(model)
        LOGGER.info(f"MODEL:\n{line_numbers_model}")
        return model

    def _enumerate_model(self, model: str):
        lines = model.split("\n")
        leading_zeros = floor(log(len(lines), 10)) + 1
        line_numbers_model = "\n".join(
            [
                f"{str(number).zfill(leading_zeros)}    {line}"
                for number, line in enumerate(lines, start=1)
            ]
        )
        return line_numbers_model

```

edvgui:
# Description

Unless specified otherwise (`no_dedent` parameter), the model provided to `project.compile` will be un-indented (we remove any space of tab at the beginning of lines if they are common to all the lines).

Before compiling the model, we log it (debug level), with line numbers.  (Makes it easier to debug error messages containing line numbers.)

closes #199 

# Self Check:

- [x] Changelog entry
- [ ] ~~Sufficient test cases (reproduces the bug/tests the requested feature)~~
- [ ] ~~End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )~~

# Reviewer Checklist:

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -621,7 +621,7 @@ license: Test License
         module.Project.set(test_project)
         test_project.load()
 
-    def compile(self, main: str, export: bool = False, no_dedent: bool = False):
+    def compile(self, main: str, export: bool = False, no_dedent: bool = False) -> None:
         """
         Compile the configuration model in main. This method will load all required modules.
 
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -19,6 +19,7 @@ import glob
 import importlib
 import json
 import logging
+import math
 import os
 import shutil
 import sys
@@ -26,7 +27,6 @@ import tempfile
 import warnings
 from collections import defaultdict
 from distutils import dir_util
-from math import floor, log
 from pathlib import Path
 from textwrap import dedent
 from types import FunctionType, ModuleType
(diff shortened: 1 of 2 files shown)


pytest_inmanta/plugin.py
```diff
@@ -619,13 +621,34 @@ def _load(self) -> None:
         module.Project.set(test_project)
         test_project.load()
 
-    def compile(self, main, export=False):
+    def compile(self, main: str, export: bool = False, no_dedent: bool = False):
```
bartv, line 624:
```suggestion
    def compile(self, main: str, export: bool = False, no_dedent: bool = False) -> None:
```

When you add typing to a method (which I really encourage) you should always do the return type as well. If there is one argument not typed (including the return type) a typechecker like mypy will mark it as incomplete

bartv (changes requested):

pytest_inmanta/plugin.py
edvgui: Damn it, I always forget the None return type, thanks

bartv (approved):

sanderr (approved):
I didn't know about `textwrap.dedent` but in hindsight I've often encountered situations where it would be useful. Cool!

PR #225, opened 2021-09-15
also perform install on every compile

wouterdb:
quick fix for changes in core, may degrade performance


sanderr (approved):
This was really dumb of me. I knew there were two occurrences but I forgot to patch this one.

Issue #236, opened 2021-10-27
modules v2 - plugin loading compatibility


sanderr:
- [x] implement #76 or reflect the plugin loading changes in core some other way (see discussion on #76)
- [x] Document that both the `inmanta_plugins` fixture and `project.get_plugin_function` support v2 modules but are not required (and not the recommended approach) to import them. v2 modules (that are dependencies of the module being tested) can just be implemented top-level.
- [x] add tests for top-level imports, `inmanta_plugins` fixture and `project.get_plugin_function` for v2 modules

(based on #209)

PR #240, opened 2021-11-10
Issue/235 modules v2 basic support

### LINKED ISSUE #235: modules v2 - add basic modules v2 support
Add support for testing a v2 module:
- [x] The get_module_info function will need to support both v1 and v2 modules
- [x] If the module being tested is v2, don't copy it to libs but verify that it is installed:
  - if not installed, raise error
  - if not editable, log warning
- [x] Document setup procedure for v2 modules: `inmanta module install [-e] .` needs to be run before running the tests.
- [x] Add basic tests against v2 module.

(based on #209)

sanderr:
# Description

Added basic support and tests for running pytest-inmanta against v2 modules. For more advanced support see #236 and #241.

The tests currently set up and activate a venv with the module to be testsed installed in it. I believe this is an acceptable solution but since we've discussed this on slack you might have something else in mind. We can discuss it further if deemed appropriate.

Requires inmanta/inmanta-core#3398.

closes #235

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:


wouterdb (approved):

tests/conftest.py
```diff
@@ -37,13 +39,26 @@ def deactive_venv():
...
     sys.prefix = old_prefix
     sys.path = old_path
-    pkg_resources.working_set = pkg_resources.WorkingSet._build_master()
+    pkg_resources.working_set = old_working_set
+    # Restore PYTHONPATH
+    if "PYTHONPATH" in os.environ:
+        if old_pythonpath is not None:
+            os.environ["PYTHONPATH"] = old_pythonpath
+        else:
+            del os.environ["PYTHONPATH"]
```
arnaudsjs, line 57:
```suggestion
    if old_pythonpath is not None:
        os.environ["PYTHONPATH"] = old_pythonpath
    elif "PYTHONPATH" in os.environ:
        del os.environ["PYTHONPATH"]
```

tests/test_basic_example_v2.py
```diff
@@ -0,0 +1,146 @@
...
+        pkg_resources.get_distribution("inmanta-core").version
+    )
+except DistributionNotFound:
+    CORE_VERSION = None
+
+
+pytestmark = pytest.mark.skipif(
+    CORE_VERSION is None or CORE_VERSION < version.Version("6.dev"),
+    reason="Skipping modules v2 tests for inmanta-core<6 (pre modules v2).",
+)
```
arnaudsjs, line 53: This marker is not used anywhere?

tests/conftest.py
```diff
@@ -37,13 +39,26 @@ def deactive_venv():
...
+    if old_os_venv is not None:
+        os.environ["VIRTUAL_ENV"] = old_os_venv
```
arnaudsjs, line 59: Shouldn't we also unset it if required?

arnaudsjs (changes requested):

tests/test_basic_example_v2.py
sanderr: `pytest` itself interprets this. This is how to set a skip mark for an entire test module: https://docs.pytest.org/en/latest/how-to/skipping.html#skip-all-test-functions-of-a-class-or-module. But now that I read the documentation again I see that there might be a better way: `if condition: pytest.skip()` instead of `pytestmark = pytest.skipif(condition)`. I'll change this.

sanderr: Done

tests/conftest.py
sanderr: I guess so. I copied this from core so I'll apply the change on my PR there as well. Same for the `PYTHON_PATH` change requested above. @wouterdb rerequesting your review for this change.

sanderr: Done

wouterdb: This magic is too dark to know what is going to happen by just reading the code, so let's see what gives.

wouterdb (approved):

arnaudsjs (approved):

sanderr:
Rerequesting review for the following changes:
- compatibility with iso4
- workaround for pypa/build#405

wouterdb (approved):

tests/test_basic_example_v2.py
```diff
@@ -63,6 +63,16 @@ def testmodulev2_venv(pytestconfig) -> env.VirtualEnv:
         # set up environment
         venv: env.VirtualEnv = env.VirtualEnv(env_path=venv_dir)
         venv.init_env()
+        # workaround for pypa/build#405: unset PYTHONPATH because it's not required in this case and it triggers a bug in build
+        with open(os.path.join(venv.site_packages_dir, "sitecustomize.py"), "a") as fd:
```
arnaudsjs, line 67: You have to use `usercustomize.py` here, because some Linux distribution already set a `sitecustomize.py` by default, causing this `sitecustomize.py` to be ignored. That would break the pytest-docker tests suite.

arnaudsjs (changes requested):

tests/test_basic_example_v2.py
sanderr: This line appends to `sitecustomize.py`, if it exists, wouldn't that work?

arnaudsjs (approved):

tests/test_basic_example_v2.py
arnaudsjs: No. This append happens in the `sitecustomize.py` file of the venv. The `sitecustomize.py` I was talking about is defined in the python system library. So that trick wouldn't work.

arnaudsjs (approved):

PR #249, opened 2021-11-23
Issue/248 module state reset

### LINKED ISSUE #248: support plugin code with side effects (i.e. module-level cache)
The recent changes to pytest-inmanta's module loading (#245) don't play well with side effects on module objects that should be contained to a compile. After discussion the proposed solution is to add support for one or more cleanup hook functions on any module's Python module (`inmanta_plugins` submodules of any depth), identified by name. For now these cleanup functions would be used by pytest-inmanta before each compile (keep state after compile so it can be investigated from test cases).

# Affected modules
The following modules should be fixed after adding support to pytest-inmanta
- [x] std
- [x] lsm
- [x] terraform
- [x] connect
- [x] bics_ce

sanderr:
# Description

Adds support for custom `inmanta_reset_state` method to clean up stateful modules between compiles

Also bumped the version because #248 illustrates a breaking changes was introduced.

@bartv I'm including you not necessarily for the implementation but for the approach, the name of the function and the clarity of the documentation.

closes #248

# Self Check:

- [x] Changelog entry
- [x] End user documentation is included

# Reviewer Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,14 @@
-# v 1.7.0 (?)
+# v 2.0.0 (?)
 Changes in this release:
 - Added support for testing v2 modules.
+- Extended to be compatible with `inmanta-core>=6`
+- Added support for custom `inmanta_reset_state` method to clean up stateful modules between compiles
 - Ensure that projects are compiled using a separate venv.
 - Fixed typing issue for `filter_args` in different method of the Project class.
+
+## Breaking changes
+- pytest-inmanta now keeps `inmanta_plugins` submodules alive accross compiles. As a result, stateful modules must implement
+    custom state cleanup logic as described in the README.
 
 # v 1.6.2 (2021-08-17)
 Changes in this release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@ Changes in this release:
 - Fixed typing issue for `filter_args` in different method of the Project class.
 
 ## Breaking changes
-- pytest-inmanta now keeps `inmanta_plugins` submodules alive accross compiles. As a result, stateful modules must implement
+- pytest-inmanta now keeps `inmanta_plugins` submodules alive across compiles. As a result, stateful modules must implement
     custom state cleanup logic as described in the README.
 
 # v 1.6.2 (2021-08-17)
--- a/README.md
+++ b/README.md
@@ -134,6 +134,26 @@ A test case, to test this plugin looks like this:
   function named `plugin_name`. As such, this line tests whether `host` is returned when the plugin function
   `hostname` is called with the parameter `fqdn`.
 
+## Advanced usage
+
+Because pytest-inmanta keeps `inmanta_plugins` submodule objects alive to support top-level imports, any stateful modules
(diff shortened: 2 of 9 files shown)


bartv (approved):

wouterdb (approved):

CHANGELOG.md
```diff
@@ -1,9 +1,15 @@
...
+# v 2.0.0 (?)
 Changes in this release:
 - Added support for testing v2 modules.
+- Extended to be compatible with `inmanta-core>=6`
+- Added support for custom `inmanta_reset_state` method to clean up stateful modules between compiles
 - Ensure that projects are compiled using a separate venv.
 - Fixed typing issue for `filter_args` in different method of the Project class.
 
+## Breaking changes
+- pytest-inmanta now keeps `inmanta_plugins` submodules alive accross compiles. As a result, stateful modules must implement
```
arnaudsjs, line 10:
```suggestion
- pytest-inmanta now keeps `inmanta_plugins` submodules alive across compiles. As a result, stateful modules must implement
```

README.md
```diff
@@ -134,6 +134,26 @@ A test case, to test this plugin looks like this:
...
+compiles. Pytest-inmanta expects such cleanup functions to be synchronous functions that live in the top-level scope (defined
+on the module object, not in a class) of a `inmanta_plugins` submodule (of any depth). Their name should start with
+"inmanta\_reset\_state" and they should not take any parameters. For example:
+
+```python
+    # <module-name>/plugins/state.py
+
+    MY_STATE = set()
+
+    def inmanta_reset_state() -> None:
```
arnaudsjs, line 150: Why is this method prefixed with `inmanta_`? Is this just to prevent name collisions in existing modules?

examples/testmodule/plugins/submod.py
```diff
@@ -5,3 +5,17 @@
...
+
+
+def inmanta_reset_state_one() -> None:
+    global MODULE_CACHE_ONE
+    MODULE_CACHE_ONE = set()
+
+
+def inmanta_reset_state_two() -> None:
+    global MODULE_CACHE_TWO
+    MODULE_CACHE_TWO = set()
```
arnaudsjs, line 21: Why support multiple reset methods?

arnaudsjs (changes requested):

README.md
sanderr: Indeed. I'm not sure about this. Do you think it would be better to remove it?

examples/testmodule/plugins/submod.py
sanderr: 1. I thought it would add more flexibility in case of complex cleanup
2. It doesn't really add complexity to support it
If you'd prefer just a single function, that would be fine by me as well.

README.md
arnaudsjs: We can keep this in. I was just wondering why this was done like that.

examples/testmodule/plugins/submod.py
arnaudsjs: We can keep it like this. I just don't see the use case, but that doesn't meant there isn't any.

arnaudsjs (approved):

PR #277, opened 2022-03-28
Issue/275 deploy all

### LINKED ISSUE #275: Add deploy_all call
We are testing a handler with inmanta pytest where many nested resources (that use the same handler) need to be deployed based on the "requires" order.

The purpose is to deploy them with pytest `project.deploy` as if they are deployed on the orchestrator.

Instead of writing project deploy in the right order:

```python
project.deploy(tenant_resource)
project.deploy(filter_resource)
```
 
We would like:
```python
project.deploy_all()
```

wouterdb:
# Description

Added deploy_all call

closes #275 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,94 @@
+# Shortcuts for various dev tasks. Based on makefile from pydantic
+.DEFAULT_GOAL := all
+isort = isort pytest_inmanta tests examples
+black = black pytest_inmanta tests examples
+
+.PHONY: install
+install:
+	pip install -U setuptools pip
+	pip install -U -r requirements.txt -r requirements.dev.txt
+	pip install -e .
+
+.PHONY: format
+format:
+	$(isort)
+	$(black)
+
+.PHONY: pep8
+pep8:
+	pip install -c requirements.txt pep8-naming flake8-black flake8-isort
+	flake8 src tests tests_common
+
+.PHONY: mypy mypy-diff mypy-save
+RUN_MYPY=MYPYPATH=stubs:src python -m mypy --soft-error-limit=-1 --html-report mypy -p pytest_inmanta
+
+mypy:
+	$(RUN_MYPY)
+
+# baseline file mypy-diff will compare to
+MYPY_BASELINE_FILE=.mypy-baseline
+# temporary file used to store most recent mypy run
+MYPY_TMP_FILE=.mypy-tmp
+# temporary file used to store baseline with line numbers filtered out
+MYPY_BASELINE_FILE_NO_LN_NB=$(MYPY_BASELINE_FILE).nolnnb
+# prepare file for diff: remove last 2 lines and filter out line numbers
+MYPY_DIFF_PREPARE=head -n -2 | sed 's/^\(.\+:\)[0-9]\+\(:\)/\1\2/'
+# read old/new line number (format +n for new or -n for old) from stdin and transform to old/new line
+MYPY_DIFF_FETCH_LINES=xargs -I{} sh -c 'sed -n -e "s/^/$(MYPY_SET_COLOUR)$$(echo {} | cut -c 1 -) /" -e "$$(echo {} | cut -c 2- -)p" $(MYPY_SELECT_FILE)'
+MYPY_SELECT_FILE=$$(if [[ "{}" == +* ]]; then echo $(MYPY_TMP_FILE); else echo $(MYPY_BASELINE_FILE); fi)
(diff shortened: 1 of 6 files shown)


Makefile
```diff
@@ -0,0 +1,94 @@
...
+# baseline file mypy-diff will compare to
+MYPY_BASELINE_FILE=.mypy-baseline
+# temporary file used to store most recent mypy run
+MYPY_TMP_FILE=.mypy-tmp
+# temporary file used to store baseline with line numbers filtered out
+MYPY_BASELINE_FILE_NO_LN_NB=$(MYPY_BASELINE_FILE).nolnnb
+# prepare file for diff: remove last 2 lines and filter out line numbers
+MYPY_DIFF_PREPARE=head -n -2 | sed 's/^\(.\+:\)[0-9]\+\(:\)/\1\2/'
+# read old/new line number (format +n for new or -n for old) from stdin and transform to old/new line
+MYPY_DIFF_FETCH_LINES=xargs -I{} sh -c 'sed -n -e "s/^/$(MYPY_SET_COLOUR)$$(echo {} | cut -c 1 -) /" -e "$$(echo {} | cut -c 2- -)p" $(MYPY_SELECT_FILE)'
```
sanderr, line 37:
```suggestion
MYPY_DIFF_FETCH_LINES=xargs -I{} bash -c 'sed -n -e "s/^/$(MYPY_SET_COLOUR)$$(echo {} | cut -c 1 -) /" -e "$$(echo {} | cut -c 2- -)p" $(MYPY_SELECT_FILE)'
```
As @FloLey found out this is not sh-compatible, it's actually bash syntax.

sanderr (approved):

pytest_inmanta/plugin.py
```diff
@@ -739,6 +793,68 @@ def deploy(
         self.finalize_handler(h)
         return ctx
 
+    def deploy_all(self, run_as_root: bool = False) -> DeployResult:
+        """
+        Deploy all resources, in the correct order
+
+        This method handles skips and failures like the normal orchestrator
```
arnaudsjs, line 800:
```suggestion
        """
        Deploy all resources, in the correct order.

        This method handles skips and failures like the normal orchestrator.
```

pytest_inmanta/plugin.py
```diff
@@ -739,6 +793,68 @@ def deploy(
...
+            )
+            if skip:
+                LOGGER.debug("Skipping %s", resource.id)
+                ctx.set_status(ResourceState.skipped)
+            else:
+                LOGGER.debug("Start executing %s", resource.id)
+                h.execute(ctx, resource)
+                LOGGER.debug("Done executing %s", resource.id)
+            self.finalize_context(ctx)
+            self.ctx = ctx
```
arnaudsjs, line 853: So `self.ctx` references the `HandlerContext` of the resource that was deployed last. Isn't that a bit weird?

arnaudsjs (changes requested):

pytest_inmanta/plugin.py
wouterdb: This is always what is has been, but I admit it is a bit weird here. Do I drop it?

arnaudsjs: I would. Leaving it in could only lead to bugs in test cases. Maybe we should set it to `None` explicitly after an invocation of `deploy_all()`

Makefile
```diff
@@ -0,0 +1,94 @@
...
+	rm -f .coverage.*
+	rm -rf build
+	find -name .env | xargs rm -rf
+	python setup.py clean
+	make -C docs clean
+
+.PHONY: docs
+docs:
+	make -C docs html
+
```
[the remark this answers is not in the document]
wouterdb: done

Makefile
```diff
@@ -0,0 +1,94 @@
...
+	rm -rf .mypy_cache
+	rm -rf mypy
+	rm -rf coverage
+	rm -rf *.egg-info
+	rm -f .coverage
+	rm -f .coverage.*
+	rm -rf build
+	find -name .env | xargs rm -rf
+	python setup.py clean
+	make -C docs clean
```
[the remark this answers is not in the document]
wouterdb: done

pytest_inmanta/plugin.py
wouterdb: ok!

arnaudsjs (approved):

PR #285, opened 2022-04-25
[FEAT] Make sure sys.executable points to the same python interpretor…

### LINKED ISSUE #280: Handlers get developer's venv in sys.executable
# Description
The value that any code running in the handler context will see in `sys.executable` in the context of a test case is not consistent with the value that would be seen in the context of a real agent.

What is expected to be set there, is the path to the python executable in the venv of the agent.  While, in the context of a test case, this will actually be set to the development's venv.  

## Background story
In the yang handler, we rely on ncdiff.  Ncdiff relies itself on pyang, which it uses by starting a process that execute the yang entrypoint installed in our path.   https://github.com/CiscoTestAutomation/ncdiff/blob/master/src/ncdiff/model.py#L662  

When an agent uses the yang module, it will install a specific version of pyang, as pinned in the `requirements.txt` file of the module.  To ensure that the pyang binary can be found, the module handlers will add to the current `PATH` environment variable, the path to the current executable's env, that it accesses using `sys.executable`.  In the context of a normal agent, it works fine.  

When creating test cases for yang-based modules, we install pytest-inmanta-yang, and with it, comes pyang.  The version of pyang installed might be different than the one pinned in the yang module.  When running test cases, the compiler will install the yang module, and all its dependencies in another venv, to avoid altering the one of the developer, but it doesn't change the value of `sys.executable`, so any logic in a handler that assumes that it points to the executable in the virtual environment of the "non-existing agent", is mistaking.  In our case, it will result in ncdiff executing the version of pyang installed in the developer's venv, instead of the one installed in the module's venv.

## Proposed fix
Set the value of `sys.executable` in the project fixture, so that it points to the exacutable in the venv of the compiler, rather than the one of the developer.

## Ongoing discussion
https://inmanta.slack.com/archives/CKRF0C8R3/p1648818348701929

Hugo-Inmanta:
… as the one used by the compiler

# Description

Override the `sys.executable` variable with the path to the python interpreter used by the compiler.

closes #280 

# Self Check:

- [x] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,7 @@
-# v 2.2.0 (?)
+# v 2.1.1 (?)
 Changes in this release:
+- The `project` fixture now makes `sys.executable` point to the compiler's executable
+
 
 # v 2.1.0 (2022-03-30)
 Changes in this release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,4 @@
-# v 2.1.1 (?)
+# v 2.2.0 (?)
 Changes in this release:
 - The `project` fixture now makes `sys.executable` point to the compiler's executable
 
--- a/examples/testhandler/tests/test_handler.py
+++ b/examples/testhandler/tests/test_handler.py
@@ -142,7 +142,7 @@ def test_close_cache(project):
     assert len(handler.cache.cache) == 0
 
 
-def test_sys_executable(project):
+def test_280_sys_executable(project):
     """
     Make sure the current python interpreter is the same as the one used by the compiler
     """
--- a/examples/testhandler/tests/test_handler.py
+++ b/examples/testhandler/tests/test_handler.py
@@ -15,6 +15,9 @@
 
     Contact: code@inmanta.com
 """
+import sys
+from pathlib import Path
+
 import pytest
 
 from inmanta import const
@@ -137,3 +140,10 @@ def test_close_cache(project):
     versions = handler.cache.counterforVersion.keys()
     assert len(versions) == 0
     assert len(handler.cache.cache) == 0
+
+
+def test_sys_executable(project):
+    """
+    Make sure the current python interpreter is the same as the one used by the compiler
+    """
+    assert Path(project._env_path) == Path(sys.executable).parent.parent
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -170,7 +170,6 @@ def project(
     DATA.clear()
     project_shared.clean()
     project_shared.init(capsys)
-    project_shared.synchronize_interpreters()
     yield project_shared
     project_shared.clean()
 
@@ -188,7 +187,6 @@ def project_no_plugins(
     DATA.clear()
     project_shared_no_plugins.clean()
     project_shared_no_plugins.init(capsys)
-    project_shared_no_plugins.synchronize_interpreters()
     yield project_shared_no_plugins
     project_shared_no_plugins.clean()
 
@@ -662,7 +660,7 @@ class Project:
         self._handlers: typing.Set[ResourceHandler] = set()
         config.Config.load_config()
 
-    def synchronize_interpreters(self) -> None:
+    def _synchronize_interpreters(self) -> None:
         """
         Store the python interpreter used by the compiler in sys.executable
         """
@@ -688,6 +686,7 @@ class Project:
         self.ctx = None
         self._handlers = set()
         self._load()
+        self._synchronize_interpreters()
         config.Config.load_config()
 
     def _create_project_and_load(self, model: str) -> module.Project:
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -660,7 +660,7 @@ class Project:
         self._handlers: typing.Set[ResourceHandler] = set()
         config.Config.load_config()
 
-    def _synchronize_interpreters(self) -> None:
+    def _set_sys_executable(self) -> None:
         """
         Store the python interpreter used by the compiler in sys.executable
         """
@@ -686,7 +686,7 @@ class Project:
         self.ctx = None
         self._handlers = set()
         self._load()
-        self._synchronize_interpreters()
+        self._set_sys_executable()
         config.Config.load_config()
 
     def _create_project_and_load(self, model: str) -> module.Project:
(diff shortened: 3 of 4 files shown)


examples/testhandler/tests/test_handler.py
```diff
@@ -137,3 +140,10 @@ def test_close_cache(project):
     versions = handler.cache.counterforVersion.keys()
     assert len(versions) == 0
     assert len(handler.cache.cache) == 0
+
+
+def test_sys_executable(project):
```
sanderr, line 145:
```suggestion
def test_280_sys_executable(project):
```
The motivation for this test is not trivial, I think linking to the ticket is useful.

pytest_inmanta/plugin.py
```diff
@@ -659,6 +662,12 @@ def __init__(
         self._handlers: typing.Set[ResourceHandler] = set()
         config.Config.load_config()
 
+    def get_compiler_executable(self) -> str:
+        if sys.platform != "win32":
+            return os.path.join(self._env_path, "bin", "python")
```
sanderr, line 667: I think on some systems `python` does not exist, only `python3`. On core we have `python_name: str = os.path.basename(sys.executable)`. Full snippet:
```
    @classmethod
    def get_python_path_for_env_path(cls, env_path: str) -> str:
        """
        For the given venv directory (`env_path`) return the path to the Python interpreter.
        """
        python_name: str = os.path.basename(sys.executable)
        return (
            os.path.join(env_path, "Scripts", python_name)
            if sys.platform == "win32"
            else os.path.join(env_path, "bin", python_name)
        )
```

examples/testhandler/tests/test_handler.py
```diff
@@ -137,3 +140,10 @@ def test_close_cache(project):
...
+    """
+    Make sure the current python interpreter is the same as the one used by the compiler
+    """
+    assert Path(project._env_path) == Path(sys.executable).parent.parent
```
sanderr, line 149: Looks good to me. You've verified that this test failed without your changes?

sanderr (changes requested):

examples/testhandler/tests/test_handler.py
Hugo-Inmanta: Yes I did :+1:

pytest_inmanta/plugin.py
sanderr: That as well. If we also use `os.path.basename(sys.executable)` here that should be fine. Or is that not what you mean?

arnaudsjs (changes requested):

pytest_inmanta/plugin.py
Hugo-Inmanta: What I meant was this:
`os.path.basename(sys.executable)` returns `python3`
But in the compiler's venv there is no `python3` , only 'python' 
And so when we return `os.path.join(env_path, "bin", python_name)`, this path doesn't actually exist.

sanderr: I think venvs always have all levels of qualification so this wouldn't be an issue. I'm not completely sure of this but it is definitely true on my system:
```
sander@bedevere:~/downloads$ rm -rf env && python -m venv env && ls env/bin/python*
env/bin/python  env/bin/python3  env/bin/python3.10
```

sanderr: On a Rocky 8 host, which does not have `python`, only `python3`:
```
[rocky@ci-pytest-inmanta-lsm-iso6-dev ~]$ rm -rf env && python3 -m venv env && ls env/bin/python*
env/bin/python  env/bin/python3  env/bin/python3.9
```

Hugo-Inmanta: Ooh, right and `python3` and `python3.9` symlink to `python`

sanderr:
```
sander@bedevere:~$ ll downloads/env/bin/python*
lrwxrwxrwx 1 sander sander 15 Apr 25 13:40 downloads/env/bin/python -> /usr/bin/python*
lrwxrwxrwx 1 sander sander  6 Apr 25 13:40 downloads/env/bin/python3 -> python*
lrwxrwxrwx 1 sander sander  6 Apr 25 13:40 downloads/env/bin/python3.10 -> python*
```
Not entirely, but semantically the same.

pytest_inmanta/plugin.py
```diff
@@ -186,6 +188,7 @@ def project_no_plugins(
     DATA.clear()
     project_shared_no_plugins.clean()
     project_shared_no_plugins.init(capsys)
+    project_shared_no_plugins.synchronize_interpreters()
```
arnaudsjs, line 191: Let's move this call into the `init()` method. Then the `init()` and the `clean()` method contain complementary logic.

arnaudsjs (changes requested):

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
-# v 2.2.0 (?)
+# v 2.1.1 (?)
```
arnaudsjs, line 1:
```suggestion
# v 2.1.1 (?)
```
```suggestion
# v 2.2.0 (?)
```

arnaudsjs, line 1: `pytest-inmanta` uses the old changelog format. You should never update the version number on such a changelog. This is done by the release pipeline. Just adding the entry to the top-level section is sufficient.
```suggestion
# v 2.2.0 (?)
```

arnaudsjs, line 1: `pytest-inmanta` uses the old changelog format. You should never update the version number on such a changelog. This is done by the release pipeline. Just adding the entry to the top-level section is sufficient.
```suggestion
# v 2.2.0 (?)
```

pytest_inmanta/plugin.py
```diff
@@ -659,6 +660,18 @@ def __init__(
         self._handlers: typing.Set[ResourceHandler] = set()
         config.Config.load_config()
 
+    def _synchronize_interpreters(self) -> None:
```
arnaudsjs, line 663: I don't really like this name. `_set_sys_executable` is a lot more clear in my opinion.

arnaudsjs, line 663: I don't really like this name. `_set_sys_executable` is a lot more clear in my opinion.

arnaudsjs (changes requested):

arnaudsjs (approved):

Issue #189, opened 2021-04-20
Add support for changing state_dir to a writable location by the current user.


edvgui:
When testing a handler which creates any file in the Inmanta state directory (by default, `var/lib/inmanta`), if the folder is not owned by the same user as the one executing the tests, the handler will fail to execute properly.  The easy fix is to make that folder owned by the same user as the one executing the tests, though that is not a clean option, and requires manual setup for everyone running the tests.

In the yang module, there is a fixture to change that state dir to a writable location for the current user, this tickets propose to port it back to pytest-inmanta.

The fixture mentioned above is the following: https://code.inmanta.com/solutions/modules/terraform/-/merge_requests/4/diffs?commit_id=74f5566e14e55c2dff05601f8f377d638d13c821

PR #297, opened 2022-06-01
Issue/189 add support for changing state dir to a writable location by the current user

Hugo-Inmanta:
# Description

Add support for changing state directory to a writable location by the current user.

part of #189 

# Self Check:

- [x] Changelog entry
~~- [ ] Sufficient test cases (reproduces the bug/tests the requested feature)~~

# Reviewer Checklist:

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -1146,11 +1146,15 @@ license: Test License
 
 @pytest.fixture(scope="function")
 def inmanta_state_dir(tmpdir_factory: "TempdirFactory") -> Iterator[str]:
+    """
+    This fixture can be overridden in the conftest of any individual project
+    in order to set the Inmanta state directory at the desired level.
+    """
     inmanta_state_dir = tmpdir_factory.mktemp("inmanta_state_dir")
     yield str(inmanta_state_dir)
-    inmanta_state_dir.remove(ignore_errors=True)
+    inmanta_state_dir.remove()
 
 
 @pytest.fixture
 def set_inmanta_state_dir(inmanta_state_dir) -> None:
-    inmanta_config.state_dir.set(str(inmanta_state_dir))
+    inmanta_config.state_dir.set(inmanta_state_dir)
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -57,7 +57,15 @@ from inmanta.protocol import json_encode
 from inmanta.resources import Resource
 
 if typing.TYPE_CHECKING:
+    # Local type stub for mypy that works with both pytest < 7 and pytest >=7
+    # https://docs.pytest.org/en/7.1.x/_modules/_pytest/legacypath.html#TempdirFactory
+    import py
     from inmanta.agent.io.local import IOBase
+
+    class TempdirFactory:
+        def mktemp(self, path: str) -> py.path.local:
+            ...
+
 
 from .handler import DATA
 from .parameters import (
@@ -1134,6 +1142,13 @@ license: Test License
     def finalize_all_handlers(self) -> None:
         for handler_instance in self._handlers:
             self.finalize_handler(handler_instance)
+
+
+@pytest.fixture(scope="function")
+def inmanta_state_dir(tmpdir_factory: "TempdirFactory") -> Iterator[str]:
+    inmanta_state_dir = tmpdir_factory.mktemp("inmanta_state_dir")
+    yield str(inmanta_state_dir)
+    inmanta_state_dir.remove(ignore_errors=True)
 
 
 @pytest.fixture
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -1152,6 +1152,6 @@ def session_temp_dir(tmpdir_factory: "TempdirFactory") -> Generator[str, None, N
 
 
 @pytest.fixture
-def use_session_temp_dir(session_temp_dir: str) -> Generator[str, None, None]:
+def use_session_temp_dir(session_temp_dir: str) -> Iterator[str]:
     inmanta_config.state_dir.set(str(session_temp_dir))
     yield inmanta_config.state_dir.get()
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -34,7 +34,7 @@ from itertools import chain
 from pathlib import Path
 from textwrap import dedent
 from types import FunctionType, ModuleType
-from typing import Dict, Generator, Iterator, List, Optional, Set, Tuple
+from typing import Dict, Iterator, List, Optional, Set, Tuple
 
 import pydantic
 import pytest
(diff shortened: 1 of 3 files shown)


pytest_inmanta/plugin.py
```diff
@@ -1130,3 +1142,16 @@ def finalize_handler(self, handler: ResourceHandler) -> None:
     def finalize_all_handlers(self) -> None:
         for handler_instance in self._handlers:
             self.finalize_handler(handler_instance)
+
+
+@pytest.fixture(scope="session")
+def session_temp_dir(tmpdir_factory: "TempdirFactory") -> Generator[str, None, None]:
```
arnaudsjs, line 1148: A shorter notation in case the  `SendType` and `ReturnType` are `None`:
```suggestion
def session_temp_dir(tmpdir_factory: "TempdirFactory") -> Iterator[str]:
```

pytest_inmanta/plugin.py
```diff
@@ -1130,3 +1142,16 @@ def finalize_handler(self, handler: ResourceHandler) -> None:
...
+    session_temp_dir = tmpdir_factory.mktemp("session")
+    yield str(session_temp_dir)
+    session_temp_dir.remove(ignore_errors=True)
+
+
+@pytest.fixture
+def use_session_temp_dir(session_temp_dir: str) -> Generator[str, None, None]:
```
arnaudsjs, line 1155:
```suggestion
def use_session_temp_dir(session_temp_dir: str) -> Iterator[str]:
```

arnaudsjs, line 1155:
```suggestion
def use_session_temp_dir(session_temp_dir: str) -> Iterator[str]:
```

arnaudsjs, line 1155: * Why do we keep the same `state_dir` for the entire session. Couldn't that cause issues?
* I think this implementation could cause issues for the `lsm` module. The `lsm` module often uses the `project` fixture of `pytest-inmanta` together with the `server` fixture of `pytest-inmanta-extensions`. Both fixtures set the `state_dir` (See: https://github.com/inmanta/inmanta-core/blob/ed50df1bbd19f28ee2716bec1272d4237b5e0f92/tests/conftest.py#L675). The following scenario might happen for a test case that uses the `server` and the `project` fixture:
    * The server fixture activates, it sets the state_dir and it starts the server
    * The project fixture activates and it sets the state_dir to a new value as well
    * Depending on the implementation of the server, part of the state is now written to the state_dir set by the server fixture and some other part to the state_dir set by the project fixture. I don't think that's desired. I am going to add @wouterdb as a reviewer as well to get his opinion on this. One possible solution would be to only update the state_dir when it's current value matches the default value. That would work, but it's still not an elegant solution.

arnaudsjs, line 1155: * Why do we keep the same `state_dir` for the entire session. Couldn't that cause issues?
* I think this implementation could cause issues for the `lsm` module. The `lsm` module often uses the `project` fixture of `pytest-inmanta` together with the `server` fixture of `pytest-inmanta-extensions`. Both fixtures set the `state_dir` (See: https://github.com/inmanta/inmanta-core/blob/ed50df1bbd19f28ee2716bec1272d4237b5e0f92/tests/conftest.py#L675). The following scenario might happen for a test case that uses the `server` and the `project` fixture:
    * The server fixture activates, it sets the state_dir and it starts the server
    * The project fixture activates and it sets the state_dir to a new value as well
    * Depending on the implementation of the server part of the state is now written to the state_dir set by the server fixture and some other part to the state_dir set by the project fixture. I don't think that's desired. I am going to add @wouterdb as a reviewer as well to get his opinion on this. One possible solution would be to only update the state_dir when it's current value matches the default value. That would work, but it's still not an elegant solution.

arnaudsjs (changes requested):

pytest_inmanta/plugin.py
wouterdb, line 1155: I don't think this is a very good name for this method. `use_session_temp_dir` doesn't very well describe what the session temp dir is used for.

wouterdb: Scoping for this one has always been difficult. Some test want to share this folder (e.g. yang test need this cache) others will want it clean.

It might be better to have a function scope 'inmanta_state_dir` fixture here, that returns the path used, that can be overridden in the conftest of any individual project, with clear documentation of how and when to do it. 

Then integrate the activation of this into the project fixture itself?

wouterdb (changes requested):

pytest_inmanta/plugin.py
```diff
@@ -1130,3 +1142,15 @@ def finalize_handler(self, handler: ResourceHandler) -> None:
...
+@pytest.fixture(scope="function")
+def inmanta_state_dir(tmpdir_factory: "TempdirFactory") -> Iterator[str]:
+    inmanta_state_dir = tmpdir_factory.mktemp("inmanta_state_dir")
+    yield str(inmanta_state_dir)
+    inmanta_state_dir.remove(ignore_errors=True)
+
+
+@pytest.fixture
+def set_inmanta_state_dir(inmanta_state_dir) -> None:
+    inmanta_config.state_dir.set(str(inmanta_state_dir))
```
arnaudsjs, line 1156:
```suggestion
    inmanta_config.state_dir.set(inmanta_state_dir)
```

arnaudsjs (changes requested):

wouterdb (approved):

pytest_inmanta/plugin.py
```diff
@@ -136,7 +146,9 @@ def project(
 
 @pytest.fixture()
 def project_no_plugins(
-    project_shared_no_plugins: "Project", capsys: "CaptureFixture"
+    project_shared_no_plugins: "Project",
+    capsys: "CaptureFixture",
+    set_inmanta_state_dir,
```
arnaudsjs, line 151:
```suggestion
    set_inmanta_state_dir: None,
```

pytest_inmanta/plugin.py
```diff
@@ -125,7 +135,7 @@ def inmanta_plugins(
 
 @pytest.fixture()
 def project(
-    project_shared: "Project", capsys: "CaptureFixture"
+    project_shared: "Project", capsys: "CaptureFixture", set_inmanta_state_dir
```
arnaudsjs, line 138:
```suggestion
    project_shared: "Project", capsys: "CaptureFixture", set_inmanta_state_dir: None
```

pytest_inmanta/plugin.py
```diff
@@ -1130,3 +1142,19 @@ def finalize_handler(self, handler: ResourceHandler) -> None:
...
+    This fixture can be overridden in the conftest of any individual project
+    in order to set the Inmanta state directory at the desired level.
+    """
+    inmanta_state_dir = tmpdir_factory.mktemp("inmanta_state_dir")
+    yield str(inmanta_state_dir)
+    inmanta_state_dir.remove()
+
+
+@pytest.fixture
+def set_inmanta_state_dir(inmanta_state_dir) -> None:
```
arnaudsjs, line 1159:
```suggestion
def set_inmanta_state_dir(inmanta_state_dir: str) -> None:
```

arnaudsjs (changes requested):

arnaudsjs (approved):

PR #302, opened 2022-06-15
Issue/3470 unserializable items in the changes set make the crud handler get stuck in updating

Hugo-Inmanta:
# Description

Ensure logs can be serialized in finalize_context

Part of [#3470](https://github.com/inmanta/inmanta-core/issues/3470)

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -735,7 +735,10 @@ class Project:
 
     def finalize_context(self, ctx: handler.HandlerContext) -> None:
         # ensure logs can be serialized
-        json_encode({"message": ctx.logs})
+        try:
+            json_encode({"message": ctx.logs})
+        except TypeError as e:
+            raise Exception(f"Unserializable logs. : {ctx.logs}") from e
 
     def get_resource(
         self, resource_type: str, **filter_args: object
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v 2.4.0 (?)
 Changes in this release:
 - Add fixture to change the Inmanta state dir to a writable location for the current user.
+- Ensure logs can be serialized in finalize_context
 
 # v 2.3.3 (2022-05-18)
 Changes in this release:


Hugo-Inmanta:
I'm not sure how to write a relevant test case for this.

pytest_inmanta/plugin.py
```diff
@@ -735,7 +735,10 @@ def get_handler(self, resource: Resource, run_as_root: bool) -> ResourceHandler:
 
     def finalize_context(self, ctx: handler.HandlerContext) -> None:
         # ensure logs can be serialized
-        json_encode({"message": ctx.logs})
+        try:
+            json_encode({"message": ctx.logs})
```
arnaudsjs, line 739: Why do we need this change? Is it just to improve the error reporting to the user? In that case the changelog entry doesn't really describe the nature of the change correctly.

arnaudsjs (changes requested):

pytest_inmanta/plugin.py
Hugo-Inmanta: The ticket mentions this:
>3. we should update pytest-inmanta to check in finilize context of invalid items are pressent.

This is how I understood it: raise an exception if somehow some unserializable items are still present in the logs. It's possible I misinterpreted it though.

Hugo-Inmanta: Oh I think I see. So since the check is already there I can close this ?

arnaudsjs: I think so. But I am not entirely sure about the intention here. The fix on inmanta-core will also ensure that no exception is raised when a non-serializable item is used. I think it would be good to cross check this with Wouter.

wouterdb: ticket #3470 is not about logs, but about the changes dict. 

If the changes in core make sure it can never be wrong, nothing is required here.

Hugo-Inmanta:
Closing this in favor of this PR: https://github.com/inmanta/inmanta-core/pull/4391

PR #323, opened 2022-09-02
[DOC] escape angle brackets so that they are correctly displayed

Hugo-Inmanta:
# Description

Escape angle brackets so that they are correctly displayed


sanderr (approved):
Nice catch!

Issue #318, opened 2022-08-22
port venv cleanup fixture fixes from core


sanderr:
- [x] inmanta/inmanta-core#4692
- [x] inmanta/inmanta-core#4717
- [x] port relevant fixes to `pytest-inmanta-lsm` as well
- [x] release both packages? (decode fix from inmanta/pytest-inmanta-lsm#243 should be merged before release)

PR #326, opened 2022-09-06
Issue/318 port venv cleanup fixture fixes from core

Hugo-Inmanta:
# Description

port venv cleanup fixture fixes from core

part of https://github.com/inmanta/pytest-inmanta/issues/318

# Self Check:

- [ ] Changelog entry

# Reviewer Checklist:

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -46,6 +46,7 @@ def deactive_venv():
     old_os_path = os.environ.get("PATH", "")
     old_prefix = sys.prefix
     old_path = sys.path
+    old_meta_path = sys.meta_path.copy()
     old_pythonpath = os.environ.get("PYTHONPATH", None)
     old_os_venv: Optional[str] = os.environ.get("VIRTUAL_ENV", None)
     old_working_set = pkg_resources.working_set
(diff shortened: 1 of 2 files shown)


tests/conftest.py
```diff
@@ -41,11 +41,13 @@ def set_cwd(testdir):
     pytest_inmanta.plugin.CURDIR = os.getcwd()
 
 
-@pytest.fixture(scope="function", autouse=True)
+@pytest.fixture(scope="function")
```
arnaudsjs, line 44: I don't think we can remove the autouse here, can we?

tests/test_basic_example_v2.py
```diff
@@ -47,7 +47,7 @@ def testmodulev2_venv(pytestconfig) -> Iterator[env.VirtualEnv]:
 
 @pytest.fixture(scope="function")
 def testmodulev2_venv_active(
-    deactive_venv: None,
+    deactive_venv: Iterator[None],
```
arnaudsjs, line 50: I don't think this is correct. The return type of the fixture is abc.Iterator[None], but the value of the argument here is None, isn't it?

arnaudsjs (changes requested):

tests/test_basic_example_v2.py
Hugo-Inmanta: Right, indeed :+1:

tests/conftest.py
Hugo-Inmanta: I naively thought that was part of the fixes that needed to be ported, I need to think about this some more.

arnaudsjs: In inmanta/inmanta-core##4717 the autouse was removed from the `deactive_venv` fixture because in the new implementation the `deactive_venv` fixture is called from the `clean_reset` which is an autouse fixture. So the `deactive_venv` fixture becomes autouse by indirection.

Hugo-Inmanta: Oh I completely missed that, I get it now

arnaudsjs (approved):

PR #359, opened 2023-01-26
Added dryrun_all and dryrun_and_deploy_all

### LINKED ISSUE #358: Add option to dryrun all of the resources to the project fixture
Also make a function that dryruns, deploys and dryruns again and does some basic sanity checks.
This would improve test writing as it will eliminate a lot of boilerplate code.

jptrindade:
# Description

Added a function to dryrun every resource and do some sanity checks.
Added another function to do a dryrun followed by a deploy and another dryrun. 
These workflows are used a lot in testing and these functions would eliminate a lot of boilerplate code.

Still missing some documentation

closes #358

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,6 @@
 Changes in this release:
 - Add `dryrun_all` method to  the `project` fixture. Does a dryrun on every resource of a project. Also does some sanity checks.
 - Add `dryrun_and_deploy_all` method to the `project` fixture. Dryruns, deploys and does a final dryrun on every resource of a project. Also does some sanity checks.
-These functions will help to write tests as they will reduce a lot of boilerplate code.
 # v 2.5.0 (2023-01-20)
 Changes in this release:
 - Fix bug where the temporary directory used to store the Inmanta project is not cleaned up when an exception occurs in the setup stage of the project\_factory fixture.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # v 2.6.0 (?)
 Changes in this release:
-
+- Add `dryrun_all` method to  the `project` fixture. Does a dryrun on every resource of a project. Also does some sanity checks.
+- Add `dryrun_and_deploy_all` method to the `project` fixture. Dryruns, deploys and does a final dryrun on every resource of a project. Also does some sanity checks.
+These functions will help to write tests as they will reduce a lot of boilerplate code.
 # v 2.5.0 (2023-01-20)
 Changes in this release:
 - Fix bug where the temporary directory used to store the Inmanta project is not cleaned up when an exception occurs in the setup stage of the project\_factory fixture.
--- a/pytest_inmanta/module/init.cf
+++ b/pytest_inmanta/module/init.cf
@@ -21,12 +21,16 @@ entity Resource extends std::PurgeableResource:
         A test resource for unit and integration testing purposes.
 
         :param skip: When set to true the handler will raise a skipresource
-        :param fail: When set to true the handler will raise an exception to mark a failure
+        :param fail: When set to true the handler will raise an exception to mark a failure whenever it performs a CRUD
+            operation.
+        :param fail_deploy: When set to true the handler will raise an exception to mark a failure when it actually performs
+            a deploy operation, i.e. create, update, delete.
     """
     string name
     string desired_value
     bool skip=false
     bool fail=false
+    bool fail_deploy=false
     string agent="internal"
 end
 
@@ -34,6 +38,27 @@ index Resource(name)
 
 implement Resource using std::none
 
+entity WrongDiffResource extends std::PurgeableResource:
+    """
+        A test resource with a faulty calculate_diff. 
+
+        :param skip: When set to true the handler will raise a skipresource
+        :param fail: When set to true the handler will raise an exception to mark a failure whenever it performs a CRUD
+            operation.
+        :param fail_deploy: When set to true the handler will raise an exception to mark a failure when it actually performs
+            a deploy operation, i.e. create, update, delete.
+    """
+    string name
+    string desired_value
+    bool skip=false
+    bool fail=false
+    bool fail_deploy=false
+    string agent="internal"
+end
+
+index WrongDiffResource(name)
+
+implement WrongDiffResource using std::none
 
 entity IgnoreResource extends std::PurgeableResource:
     """
--- a/pytest_inmanta/module/init.cf
+++ b/pytest_inmanta/module/init.cf
@@ -25,12 +25,15 @@ entity Resource extends std::PurgeableResource:
             operation.
         :param fail_deploy: When set to true the handler will raise an exception to mark a failure when it actually performs
             a deploy operation, i.e. create, update, delete.
+        :param wrong_diff: When set to true the handler will return a wrong diff result.
+            Causes dryrun_and_deploy_all to fail on the last dryrun.
     """
     string name
     string desired_value
     bool skip=false
     bool fail=false
     bool fail_deploy=false
+    bool wrong_diff=false
     string agent="internal"
 end
 
@@ -38,27 +41,6 @@ index Resource(name)
 
 implement Resource using std::none
 
-entity WrongDiffResource extends std::PurgeableResource:
-    """
-        A test resource with a faulty calculate_diff. 
-
-        :param skip: When set to true the handler will raise a skipresource
-        :param fail: When set to true the handler will raise an exception to mark a failure whenever it performs a CRUD
-            operation.
-        :param fail_deploy: When set to true the handler will raise an exception to mark a failure when it actually performs
-            a deploy operation, i.e. create, update, delete.
-    """
-    string name
-    string desired_value
-    bool skip=false
-    bool fail=false
-    bool fail_deploy=false
-    string agent="internal"
-end
-
-index WrongDiffResource(name)
-
-implement WrongDiffResource using std::none
 
 entity IgnoreResource extends std::PurgeableResource:
     """
--- a/pytest_inmanta/module/init.py
+++ b/pytest_inmanta/module/init.py
@@ -16,6 +16,8 @@
     Contact: code@inmanta.com
 """
 
+import typing
+
 from inmanta import resources
 from inmanta.agent import handler
 from pytest_inmanta.handler import DATA
@@ -25,7 +27,12 @@ KEY_PREFIX = "unittest_"
 
 @resources.resource("unittest::Resource", id_attribute="name", agent="agent")
 class Resource(resources.PurgeableResource):
-    fields = ("name", "desired_value", "skip", "fail")
+    fields = ("name", "desired_value", "skip", "fail", "fail_deploy")
+
+
+@resources.resource("unittest::WrongDiffResource", id_attribute="name", agent="agent")
+class WrongDiffResource(resources.PurgeableResource):
+    fields = ("name", "desired_value", "skip", "fail", "fail_deploy")
 
 
 @handler.provider("unittest::Resource", name="test")
@@ -34,8 +41,9 @@ class ResourceHandler(handler.CRUDHandler):
         self, ctx: handler.HandlerContext, resource: resources.PurgeableResource
     ) -> None:
         ctx.info(
-            "Resource fail %(fail)s skip %(skip)s",
+            "Resource fail %(fail)s fails deploy %(fail_deploy)s skip %(skip)s",
             fail=resource.fail,
+            fail_deploy=resource.fail_deploy,
             skip=resource.skip,
         )
 
@@ -52,6 +60,9 @@ class ResourceHandler(handler.CRUDHandler):
                 raise handler.InvalidOperation()
 
         if resource.fail:
+            raise handler.InvalidOperation()
+
+        if resource.fail_deploy and not ctx.is_dry_run():
             raise handler.InvalidOperation()
 
         if resource.name not in DATA:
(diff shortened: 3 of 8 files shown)


jptrindade:
I'm also taking suggestions of a better name than `dryrun_and_deploy_all`

wouterdb (changes requested):

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # v 2.6.0 (?)
 Changes in this release:
-
+- Add `dryrun_all` method to  the `project` fixture. Does a dryrun on every resource of a project. Also does some sanity checks.
+- Add `dryrun_and_deploy_all` method to the `project` fixture. Dryruns, deploys and does a final dryrun on every resource of a project. Also does some sanity checks.
```
sanderr, line 4: What is the purpose of the first dryrun?

pytest_inmanta/plugin.py
```diff
@@ -951,6 +951,39 @@ def dryrun_resource(
...
+        :param run_as_root: run the mock agent as root
+        :param assert_create_or_delete: assert that every resource will either be created or deleted.
+        """
+        dryrun_result = self.dryrun_all(
+            run_as_root=run_as_root, assert_create_or_delete=assert_create_or_delete
+        )
+        deploy_result = self.deploy_all(run_as_root=run_as_root)
+
+        assert not self.dryrun_all(run_as_root=run_as_root)
+        return {"dryrun": dryrun_result, "deploy": deploy_result.results}
```
sanderr, line 985: We generally try to avoid untyped dicts for structured data. You could use a `TypedDict` but I would suggest introducing a dataclass for this result. It could also help to add some helper methods, so the end user can inspect state like e.g. `assert dry_run_and_deploy_all().all_ok()` (naming might not be good but I think it gets the point accross).

pytest_inmanta/plugin.py
```diff
@@ -951,6 +951,39 @@ def dryrun_resource(
         assert ctx.status == status
         return ctx.changes
 
+    def dryrun_all(
+        self, run_as_root: bool = False, assert_create_or_delete: bool = False
+    ) -> Dict[Resource, HandlerContext]:
+        """
+        Runs a dryrun for every resource.
```
sanderr, line 958: This only returns resources with changes. I can't comment on the correctness of that, but could you add it to the docstring so it's clear to the user?

sanderr (changes requested):
I think the tests could be expanded a bit:
- assert the method returns the expected value
- add some failure scenarios (if feasible): what if the deploy fails? What if the dryrun after the deploy is not empty?

CHANGELOG.md
jptrindade: The purpose of the first dryrun is to check if there is the `purged` key in the results. Meaning that the resource is not currently deployed and is being created. To be honest, if `assert_create_or_delete` is `False`I don't think there is much use in the first dryrun (maybe checking if it fails before the deploy?)

pytest_inmanta/plugin.py
bartv: For new things it should be dataclass for better typing of existing code use TypedDict

pytest_inmanta/plugin.py
jptrindade: I changed it to return everything. 

> I think the tests could be expanded a bit:
> 
>     * assert the method returns the expected value
> 
>     * add some failure scenarios (if feasible): what if the deploy fails? What if the dryrun after the deploy is not empty?

Regarding this comment. I tried to do more thorough tests but I couldn't actually get the dryrun to succeed and the deploy to fail. The dryrun after the deploy I assume is only empty if something went wrong with the deploy. Do you have any idea on how to test this?

wouterdb (changes requested):

pytest_inmanta/plugin.py
```diff
@@ -773,23 +812,7 @@ def finalize_context(self, ctx: handler.HandlerContext) -> None:
...
-            resource = next(
-                get_resources_matching(
-                    self.resources.values(), resource_type, **filter_args
-                )
-            )
-            resource = self.check_serialization(resource)
-            return resource
-        except StopIteration:
-            return None
+        return get_resource(self.resources.values(), resource_type, **filter_args)
```
wouterdb, line 815: I would not remove the docstring here

pytest_inmanta/plugin.py
```diff
@@ -581,15 +581,49 @@ def apply_filter(resource: Resource) -> bool:
...
+
+    try:
+        resource = next(get_resources_matching(resources, resource_type, **filter_args))
+        resource = check_serialization(resource)
+        return resource
+    except StopIteration:
+        return None
+
+
+class GenericResult:
```
wouterdb, line 607:
```suggestion
class Result:
```

pytest_inmanta/plugin.py
```diff
@@ -581,15 +581,49 @@ def apply_filter(resource: Resource) -> bool:
...
 
-    def assert_all(self, status: ResourceState = ResourceState.deployed):
+    def assert_all(self, status):
         for r, ct in self.results.items():
             assert (
                 ct.status == status
-            ), f"Resource {r.id} has deploy status {ct.status}, expected {status}"
+            ), f"Resource {r.id} has status {ct.status}, expected {status}"
+
+    def assert_has_no_changes(self):
```
wouterdb, line 617: type annotations are required on this repo, please use the PR template to see if all things are ok  https://github.com/inmanta/pytest-inmanta/blob/master/.github/pull_request_template.md

pytest_inmanta/plugin.py
```diff
@@ -951,6 +974,41 @@ def dryrun_resource(
...
+
+        if assert_create_or_delete:
+            first_dryrun.assert_resources_have_purged()
+
+        deploy = self.deploy_all(run_as_root=run_as_root)
+        deploy.assert_all(const.ResourceState.deployed)
+
+        last_dryrun = self.dryrun_all(run_as_root=run_as_root)
+        first_dryrun.assert_all(const.ResourceState.dry)
+        first_dryrun.assert_has_no_changes
```
wouterdb, line 1008:
```suggestion
        first_dryrun.assert_has_no_changes()
```

wouterdb (changes requested):
API wise, this looks very nice.

pytest_inmanta/plugin.py
```diff
@@ -581,15 +581,49 @@ def apply_filter(resource: Resource) -> bool:
...
+    def assert_has_no_changes(self) -> None:
+        for r, ct in self.results.items():
+            assert (
+                ct.change is const.Change.nochange
+            ), f"Resource {r.id} has changes {ct.changes}, expected no changes"
+
+    def assert_resources_have_purged(self) -> None:
+        for r, ct in self.results.items():
+            assert ct.changes, f"Resource {r.id} has no changes, expected some changes"
+            assert "purged" in ct.changes
```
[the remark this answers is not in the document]
sanderr: From looking at its usage I get the impression that the implementation is in fact as intended. It could really use a docstring though.

pytest_inmanta/plugin.py
```diff
@@ -951,38 +986,40 @@ def dryrun_resource(
...
+
+        if assert_create_or_delete:
+            first_dryrun.assert_resources_have_purged()
+
+        deploy = self.deploy_all(run_as_root=run_as_root)
+        deploy.assert_all(const.ResourceState.deployed)
+
+        last_dryrun = self.dryrun_all(run_as_root=run_as_root)
+        first_dryrun.assert_all(const.ResourceState.dry)
+        first_dryrun.assert_has_no_changes()
```
sanderr, line 1020: Is this what you meant?
```suggestion
        last_dryrun.assert_all(const.ResourceState.dry)
        last_dryrun.assert_has_no_changes()
```

pytest_inmanta/plugin.py
```diff
@@ -951,38 +986,40 @@ def dryrun_resource(
...
+            {
+                r: self.dryrun(r, run_as_root=run_as_root)
+                for r in self.resources.values()
+            }
+        )
 
     def dryrun_and_deploy_all(
         self, run_as_root: bool = False, assert_create_or_delete: bool = False
-    ) -> Dict[str, Dict[Resource, HandlerContext]]:
+    ) -> List[Result]:
```
sanderr, line 1003: I would make this a tuple and clarify in the docstring what the ordering means.
```suggestion
    ) -> "tuple[Result, Result, Result]"
```

pytest_inmanta/plugin.py
```diff
@@ -951,38 +986,40 @@ def dryrun_resource(
...
+        deploy = self.deploy_all(run_as_root=run_as_root)
+        deploy.assert_all(const.ResourceState.deployed)
+
+        last_dryrun = self.dryrun_all(run_as_root=run_as_root)
+        first_dryrun.assert_all(const.ResourceState.dry)
+        first_dryrun.assert_has_no_changes()
 
-        assert not self.dryrun_all(run_as_root=run_as_root)
-        return {"dryrun": dryrun_result, "deploy": deploy_result.results}
+        return [first_dryrun, deploy, last_dryrun]
```
sanderr, line 1022:
```suggestion
        return (first_dryrun, deploy, last_dryrun)
```
See other comment.

pytest_inmanta/plugin.py
sanderr: Or even better, use a dataclass
```
@dataclass
class DryrunResult:
    first_dryrun: Result
    deploy: Result
    dryrun: Result
```

You could optionally make it iterable if that's important.

pytest_inmanta/plugin.py
sanderr: Could you still use help with this?

jptrindade:
@wouterdb https://github.com/inmanta/pytest-inmanta/pull/359/commits/b5ea8f0ece3ce1a33cd71ce5640840fc4e7541b1 The main changes are in this commit.

pytest_inmanta/plugin.py
```diff
@@ -621,6 +624,10 @@ def assert_has_no_changes(self) -> None:
             ), f"Resource {r.id} has changes {ct.changes}, expected no changes"
 
     def assert_resources_have_purged(self) -> None:
+        """
+        Asserts that `purged` is in the changes. This is helpful to assert if the
+        resources are to be created (`purged` set to True) or deleted (`purged` set to False)
+        """
```
sanderr, line 630: Nice, I find this a lot more clear now.

pytest_inmanta/module/init.cf
```diff
@@ -21,19 +21,44 @@ entity Resource extends std::PurgeableResource:
...
     bool fail=false
+    bool fail_deploy=false
     string agent="internal"
 end
 
 index Resource(name)
 
 implement Resource using std::none
 
+entity WrongDiffResource extends std::PurgeableResource:
```
sanderr, line 41: Any reason you make this an entirely new resource rather than an option in the other one?

pytest_inmanta/module/init.py
```diff
@@ -54,6 +62,9 @@ def read_resource(
         if resource.fail:
             raise handler.InvalidOperation()
 
+        if resource.fail_deploy and not ctx.is_dry_run():
```
sanderr, line 65: In addition to this there is also the check `if resource.name in DATA` above. I'm not sure if we need to take that in to account as well. @wouterdb wdyt?

pytest_inmanta/plugin.py
```diff
@@ -615,6 +656,30 @@ def get_context_for(
...
+
+
+class DeployResult(Result):
+    """
+    Here for backwards compatibility reasons.
+    """
+
+
+@dataclass
+class ResultCollection:
```
sanderr, line 678:
```suggestion
class DryrunResultCollection:
```
Because it's not just any collection of results.

pytest_inmanta/module/init.cf
jptrindade: Oh that makes a lot more sense, I wanted to use a different handler so I had to make a different resource. But I can just make an option on the other one and overwrite the `calculate_diff` there

sanderr: Yes, I think that would be simpler. And in the body of `calculate_diff` just call `super().calculate_diff()` if the option is disabled.

pytest_inmanta/plugin.py
jptrindade: It is also not just any collection of dryrun results, since we actually deploy there. Maybe `DeployResultCollecton` might be better, because technically a dryrun is just a simulated deploy

sanderr: Sounds good to me.

pytest_inmanta/module/init.py
sanderr: I'm approving the PR but I would await Wouter's response before merging. The same question applies to the `wrong_diff`.

sanderr (approved):

CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
...
+These functions will help to write tests as they will reduce a lot of boilerplate code.
```
wouterdb, line 5:
```suggestion
```

pytest_inmanta/plugin.py
```diff
@@ -803,7 +858,7 @@ def deploy(
 
         assert h is not None
 
-        ctx = handler.HandlerContext(resource)
+        ctx = handler.HandlerContext(resource, dry_run=dry_run)
```
wouterdb, line 861: Nice!

pytest_inmanta/plugin.py
```diff
@@ -615,6 +656,30 @@ def get_context_for(
...
+        """
+        Get a resource of the given type and given filter on the results attributes from a Result object.
+        If multiple resource match, the first one is returned. If none match, None is returned.
+
+        :param resource_type: The exact type used in the model (no super types)
+        """
+        return get_resource(self.results.keys(), resource_type, **filter_args)
+
+
+class DeployResult(Result):
```
wouterdb, line 671:
```suggestion
DeployResult = Result
```

I think this is more robust wrt to isinstance checks ?

wouterdb (approved):

pytest_inmanta/plugin.py
sanderr: Good point, we need to be backwards compatible in two directions.

pytest_inmanta/module/init.py
wouterdb: ok, this does aan new functionality, so I retract my comment

wouterdb (approved):

PR #364, opened 2023-02-23
Issue/more project parametrization

edvgui:
# Description

- Facilitate parametrization of the metadata of a project created using the project fixture.  
- Add option to enable (when supported) the `agent_install_dependency_modules` option of a project.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -314,7 +314,7 @@ def project_factory(
             raise
 
     with open(os.path.join(project_dir, "project.yml"), "w+") as fd:
-        yaml.dump(project_metadata.dict(), fd)
+        yaml.safe_dump(json.loads(project_metadata.json()), fd)
 
     ensure_current_module_install(
         os.path.join(project_dir, "libs"),
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -70,7 +70,6 @@ if typing.TYPE_CHECKING:
 
 from .handler import DATA
 from .parameters import (
-    inm_agent_install_dependency_modules,
     inm_install_mode,
     inm_mod_in_place,
     inm_mod_repo,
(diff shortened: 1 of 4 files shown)


pytest_inmanta/plugin.py
```diff
@@ -289,17 +314,12 @@ def project_factory(
...
-        metadata: typing.Mapping[str, object] = {
-            "name": "testcase",
-            "description": "Project for testcase",
-            "repo": repos,
-            "modulepath": modulepath,
-            "downloadpath": "libs",
-            "install_mode": install_mode.value,
-        }
-        yaml.dump(metadata, fd)
+        yaml.safe_dump(json.loads(project_metadata.json()), fd)
```
edvgui, line 317: The `pydantic.BaseModel.dict()` doesn't output something that can safely be dumped as yaml (`yaml.dump` would choke on enums) so I am serializing and de-serializing it here.  Any other suggestion?

pytest_inmanta/plugin.py
```diff
@@ -289,17 +314,14 @@ def project_factory(
...
-        }
-        yaml.dump(metadata, fd)
+        # pydantic.BaseModel.dict() doesn't produce data that can be serialized
+        # so we first serialize it as json, then load it and dump it as yaml.
+        yaml.safe_dump(json.loads(project_metadata.json()), fd)
 
-    ensure_current_module_install(os.path.join(project_dir, "libs"), in_place)
+    ensure_current_module_install(
+        os.path.join(project_dir, "libs"),
+        in_place=inm_mod_in_place.resolve(request.config),
```
sanderr, line 317: Why not use the variable anymore?

pytest_inmanta/plugin.py
sanderr: Nope, looks OK to me.

sanderr (approved):
- as discussed, I think it would be better to leave out the option and just introduce the `project_metadata` fixture. It can then be easily overridden where required.
- ideally, we should add a test for this, but that can wait until next week if there are more urgent matters now.

pytest_inmanta/plugin.py
edvgui: Because it is not used in another place anymore (the other logic has been moved to the project_metadata fixture)

PR #376, opened 2023-03-28
Issue/re write state dir in project

edvgui:
# Description

The `set_inmanta_state_dir` fixture doesn't work well with the `project` fixture, as the latter will reload the config, overwriting any other value set by the `set_inmanta_state_dir` fixture.

This PR adds:
1. A test case to reproduce the issue
2. A proposal fix, but I don't really like it to be honest, any input is welcome

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/examples/testmodulev2/inmanta_plugins/testmodulev2/__init__.py
+++ b/examples/testmodulev2/inmanta_plugins/testmodulev2/__init__.py
@@ -3,9 +3,64 @@
     Contact: code@inmanta.com
     License: Apache 2.0
 """
+import pathlib
+
+import inmanta.agent.handler
+import inmanta.config
+import inmanta.resources
 from inmanta.plugins import plugin
 
 
 @plugin
 def myplugin(x: "int") -> "int":
     return x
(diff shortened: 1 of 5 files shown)


sanderr (commented):
1. I think an easy win would be to move the loading of the config and your fixes to `pytest_inmanta.plugin.ProjectLoader`. It's a "Singleton providing methods for managing project loading and assosiated side effects" to quote its docstring. That sounds like exactly what you're doing here so I think it's probably a nice fit. It's documented rather well (I think) so you should be able to find your way around it.
2. How hard would it be to make this more generic? i.e. extend it to cover all config values? Because this hardcodes the assumption that a fixture exists for this one and for this one only.
3. I think we may need a cleanup hook as well. If the state dir is set by a fixture, we want to keep its value accross compiles but not accross test cases. Unless it's a session-scoped fixture. We should consider both possibilities. I don't think `ProjectLoader` already deals with a scenario like this.

sanderr (approved):

PR #384, opened 2023-05-02
Issue/test editable installs

edvgui:
# Description

Test editable installs.  cf. https://github.com/inmanta/pytest-inmanta-lsm/pull/312#discussion_r1182216046

# Self Check:

- [x] Attached issue to pull request (no ticket)
- [x] Changelog entry (no change in pytest-inmanta itself)

### DIFF
--- a/tests/test_basic_example_v2.py
+++ b/tests/test_basic_example_v2.py
@@ -34,13 +34,26 @@ if not core.SUPPORTS_MODULES_V2:
     )
 
 
-@pytest.fixture(scope="session")
-def testmodulev2_venv(pytestconfig) -> Iterator[env.VirtualEnv]:
+@pytest.fixture(scope="function")
+def testmodulev2_editable_install() -> bool:
+    raise NotImplementedError(
+        "Tests using the testmodulev2 should have a marker specifying whether the module "
+        "should be installed in editable mode or not."
+    )
+
+
+@pytest.fixture(scope="function")
+def testmodulev2_venv(
+    testdir: pytest.Testdir,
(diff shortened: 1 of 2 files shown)


tests/test_basic_example_v2.py
sanderr, line 1: Actually, with the parameter you added to the context manager we don't even need to make changes to this file, do we? We have consistency between pytest-inmanta and pytest-inmanta lsm in the context manager, we just use it slightly differently here so we can use the session-scoped fixture?

tests/test_basic_example_v2.py
```diff
@@ -34,13 +34,26 @@
...
+        "Tests using the testmodulev2 should have a marker specifying whether the module "
+        "should be installed in editable mode or not."
+    )
+
+
+@pytest.fixture(scope="function")
+def testmodulev2_venv(
+    testdir: pytest.Testdir,
+    pytestconfig: pytest.Config,
+    testmodulev2_editable_install: bool,
```
sanderr, line 49: Wow, does this get its value from the marker of the test that requires the fixture?

tests/test_basic_example_v2.py
edvgui: We could, it just mean that we don't test modules installed in editable mode then

sanderr: Good point. For pytest-inmanta I don't think it should make a difference but it's good to cover both cases. If it doesn't have too big an impact on test time I'm all for it.

sanderr (approved):

PR #396, opened 2023-06-22
Issue/5993 add dedicated pip index url for full project

Hugo-Inmanta:
# Description

Part of https://github.com/inmanta/inmanta-core/issues/5993

# Self Check:

- [x] Changelog entry


pytest_inmanta/parameters.py
```diff
@@ -96,6 +96,14 @@
     legacy=inm_mod_repo_legacy,
 )
 
+pip_index_url = ListTestParameter(
+    argument="--pip-index-url",
+    environment_variable="INMANTA_PIP_INDEX_URL",
```
sanderr, line 101: Nice! I already have this env var set.

pytest_inmanta/parameters.py
```diff
@@ -96,6 +96,14 @@
...
+    usage=("List of project-wide pip indexes"),
```
sanderr, line 102:
```suggestion
    usage="List of project-wide pip indexes",
```
These are redundant.

sanderr (changes requested):
I'm not sure this suffices. We have to think about compatibility here. What do we want this `--pip-index-url` option for `pytest-inmanta` to represent?
1. Current implementation: it sets the `pip.index_url` config option that is only available on the latest version of core.
2. Backwards compatibility approach: it is pytest-inmanta's setting to set the pip index url: on new versions of core it sets `pip.index_url`, on older versions it falls back to adding a v2 package index. This also means we deprecate setting it through the module repo option.

I am personally in favor of 2. There is something to say for 1 (new config option -> new cli option) but I think it could be very unintuitive for an end-user.

pytest_inmanta/__init__.py
sanderr, line 1: Perhaps put this in `pytest_inmanta.core`?

pytest_inmanta/__init__.py
```diff
@@ -15,3 +15,26 @@
...
     Contact: code@inmanta.com
 """
+from typing import Optional
+
+import pkg_resources
+from pkg_resources import DistributionNotFound
+
+from packaging import version
+
+CORE_VERSION: Optional[version.Version]
```
sanderr, line 25: In the test code I would import this file instead of doing it over there.

FYI: inmanta/inmanta-core#6175

pytest_inmanta/plugin.py
```diff
@@ -56,6 +56,8 @@
 from inmanta.protocol import json_encode
 from inmanta.resources import Resource
 
+from . import SUPPORTS_PROJECT_PIP_INDEX
```
sanderr, line 59: We don't generally use relative imports.

pytest_inmanta/__init__.py
```diff
@@ -15,3 +15,26 @@
...
+try:
+    CORE_VERSION = version.Version(
+        pkg_resources.get_distribution("inmanta-core").version
+    )
+except DistributionNotFound:
+    CORE_VERSION = None
+
+
+SUPPORTS_PROJECT_PIP_INDEX: bool = (
+    CORE_VERSION is not None and CORE_VERSION >= version.Version("9.2.0.dev")
```
sanderr, line 39:
```suggestion
    CORE_VERSION is not None and CORE_VERSION >= version.Version("9.3.0.dev")
```
We will delay the merge until after the release. This means you will also need to do a minor bump on core (the release process will only make it `9.2.1`.

pytest_inmanta/plugin.py
```diff
@@ -261,19 +274,36 @@ def project_metadata(request: pytest.FixtureRequest) -> module.ProjectMetadata:
...
-        name="testcase",
-        description="Project for testcase",
-        repo=repos,
-        modulepath=modulepath,
-        downloadpath="libs",
-        install_mode=inm_install_mode.resolve(request.config).value,
-    )
+    if SUPPORTS_PROJECT_PIP_INDEX:
+        # On newer versions of core we set the pip.index_url of the project.yml file
+        pip_config: ProjectPipConfig = ProjectPipConfig(index_url=list(pip_index_urls))
```
sanderr, line 285: We should ensure backwards compatibility in the other direction as well: if the user still uses the deprecated option we should supply it to `pip.index_url`.

sanderr (changes requested):

pytest_inmanta/plugin.py
```diff
@@ -274,15 +272,22 @@ def project_metadata(request: pytest.FixtureRequest) -> module.ProjectMetadata:
...
     if SUPPORTS_PROJECT_PIP_INDEX:
         # On newer versions of core we set the pip.index_url of the project.yml file
-        pip_config: ProjectPipConfig = ProjectPipConfig(index_url=list(pip_index_urls))
+        repos_urls: List[str] = [
+            repo["url"]
+            for repo in repos
+            if repo["type"] == module.ModuleRepoType.package
+        ]
+        pip_config: ProjectPipConfig = ProjectPipConfig(
+            index_url=list(set(list(index_urls) + repos_urls))
```
sanderr, line 289: Same comment about set and insertion order here.

sanderr (changes requested):

Hugo-Inmanta:
I still have some mypy errors: 

```
+ pytest_inmanta/plugin.py:285: error: Value of type "object" is not indexable  [index]
+ pytest_inmanta/plugin.py:287: error: Value of type "object" is not indexable  [index]
+ pytest_inmanta/plugin.py:302: error: Argument "install_mode" to "ProjectMetadata" has incompatible type "Union[str, str, str]"; expected "InstallMode"  [arg-type]
+ pytest_inmanta/plugin.py:312: error: Argument "repo" to "ProjectMetadata" has incompatible type "list[object]"; expected "list[ModuleRepoInfo]"  [arg-type]
```

sanderr (changes requested):
About the mypy errors: those seem to be a side effect of the fact that we need to support multiple versions of core. Until we can clean that up I think we can leave most of it as is, because we do have pydantic validation, even if mypy doesn't recognize it. There is one part that we may need to address now: I think the current implementation may break on iso4 if `index_urls` is set because you add them as v2 source repos, and iso4 (iirc, should be checked) only accepts strings.
- typing this correctly is more difficult because it depends on the version of core, so we could consider most of that out of scope until we can clean this up. But I think it might be good to

sanderr (approved):

PR #401, opened 2023-07-04
[FIX] make new behaviour iso7+ only

Hugo-Inmanta:
# Description

As discussed [here](https://inmanta.slack.com/archives/CKRF0C8R3/p1688386956424409?thread_ts=1688132643.347189&cid=CKRF0C8R3):

> 1. on iso6, if you use --pip-index-url pytest-inmanta logs a warning that a project-wide pip index url is not supported on this version of inmanta-core and that it will use the index as a v2 package source instead.
> 2. on iso7+, stick with the current warning
> Wdyt? We should update the changelog note in that case.

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 Changes in this release:
 - Log a warning when the id_attribute of a resource is called id.
 - Ignore `__pycache__` dirs when copying the current module to the test project dir
-- Deprecate setting pip indexes through `--module-repo` option with type `package` in favour of newly introduced `--pip-index-url` option.
+- [inmanta-core > iso7] Deprecate setting pip indexes through `--module-repo` option with type `package` in favour of newly introduced `--pip-index-url` option.
 
 # v 2.7.0 (2023-02-23)
 Changes in this release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 Changes in this release:
 - Log a warning when the id_attribute of a resource is called id.
 - Ignore `__pycache__` dirs when copying the current module to the test project dir
-- [inmanta-core > iso7] Deprecate setting pip indexes through `--module-repo` option with type `package` in favour of newly introduced `--pip-index-url` option.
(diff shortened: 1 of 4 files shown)


pytest_inmanta/core.py
```diff
@@ -35,9 +35,9 @@
 except DistributionNotFound:
     CORE_VERSION = None
 
-
+# Setting a project-wide pip index is only supported for iso7+
 SUPPORTS_PROJECT_PIP_INDEX: bool = (
-    CORE_VERSION is not None and CORE_VERSION >= version.Version("9.2.0.dev")
+    CORE_VERSION is not None and CORE_VERSION >= version.Version("10.0.0.dev")
```
Hugo-Inmanta, line 40: This is the right version number for iso7+ ?

CHANGELOG.md
```diff
@@ -2,7 +2,7 @@
 Changes in this release:
 - Log a warning when the id_attribute of a resource is called id.
 - Ignore `__pycache__` dirs when copying the current module to the test project dir
-- Deprecate setting pip indexes through `--module-repo` option with type `package` in favour of newly introduced `--pip-index-url` option.
+- [inmanta-core > iso7] Deprecate setting pip indexes through `--module-repo` option with type `package` in favour of newly introduced `--pip-index-url` option.
```
Hugo-Inmanta, line 5: I'm not very satisfied with this changelog message, but technically speaking nothing has been added for iso6

sanderr: Perhaps it reads better if we invert it? "Added --pip-index-url to set corresponding project config section for inmanta-core 9.0"?

pytest_inmanta/core.py
sanderr: 9.0.0.dev

sanderr (changes requested):
Looks good apart from the version number

sanderr (approved):

sanderr:
FYI, #402

Issue #404, opened 2023-07-11
reconsider naming of index-url option


sanderr:
- [ ] Check how multiple index-urls can be specified. Is it the same as for `--module-repo` and `MODULE_REPO`? If so, we should use the singular form `--index-url` instead. The help string should also be updated to be more clear on how to use it, similar to the --module-repo help string.
- [ ] Update the deprecation warning to detect whether the option or the env var was used and update the message accordingly.
- [ ] Mention in the warning that this will also raise a warning during compilation and that that warning will disappear when this one is addressed.

sanderr:
Hey team! Please [add your planning poker estimate](https://app.zenhub.com/workspaces/core-5a7c152b04a1652024289b7b/issues/inmanta/pytest-inmanta/404?planning-poker) with Zenhub @FloLey @Hugo-Inmanta

PR #408, opened 2023-07-18
[WIP] update warning message

Hugo-Inmanta:
# Description

- [x] Check how multiple index-urls can be specified. Is it the same as for `--module-repo` and `MODULE_REPO`? If so, we should use the singular form `--index-url` instead. The help string should also be updated to be more clear on how to use it, similar to the --module-repo help string.
- [x] Update the deprecation warning to detect whether the option or the env var was used and update the message accordingly.
- [x] Mention in the warning that this will also raise a warning during compilation and that that warning will disappear when this one is addressed.

closes https://github.com/inmanta/pytest-inmanta/issues/404

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -57,6 +57,7 @@ from inmanta.export import Exporter, ResourceDict, cfg_env
 from inmanta.protocol import json_encode
 from inmanta.resources import Resource
 from pytest_inmanta.core import SUPPORTS_PROJECT_PIP_INDEX
+from .test_parameter.parameter import ValueSetBy
 
 if typing.TYPE_CHECKING:
     # Local type stub for mypy that works with both pytest < 7 and pytest >=7
@@ -80,7 +81,7 @@ from .parameters import (
     inm_no_load_plugins,
     inm_no_strict_deps_check,
     inm_venv,
-    pip_index_urls,
+    pip_index_url,
 )
 from .test_parameter import ParameterNotSetException, TestParameterRegistry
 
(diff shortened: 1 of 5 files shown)


pytest_inmanta/test_parameter/list_parameter.py
```diff
@@ -87,19 +87,28 @@ def validate(self, raw_value: object) -> Sequence[str]:
...
+                return self.validate(env_var.split(" "))
+
+            if self.default is not None:
+                value_set_using = ValueSetBy.DEFAULT_VALUE
+                return self.default
+
+            raise ParameterNotSetException(self)
+
+        finally:
+            self._value_set_using = value_set_using
```
sanderr, line 114: Why not just assign to the attribute directly?

pytest_inmanta/test_parameter/parameter.py
```diff
@@ -164,6 +165,15 @@ def pytest_addoption(cls, parser: "Parser") -> None:
...
 
+class ValueSetBy(Enum):
+    """
+    This class is used to record how the value was provided for a test parameter.
+    """
+
+    DEFAULT_VALUE: str = "DEFAULT_VALUE"
+    CLI: str = "CLI"
+    ENV_VARIABLE: str = "ENV_VARIABLE"
+    UNSET: str = "UNSET"
```
sanderr, line 176: This doesn't seem to be used, is it? I would remove it, `Optional` (as you use it now) is perfect for that.

sanderr (approved):

pytest_inmanta/test_parameter/parameter.py
Hugo-Inmanta: Indeed, this is an artifact from a previous implementation that I forgot to remove

pytest_inmanta/test_parameter/list_parameter.py
Hugo-Inmanta: I can't think of a good reason why not to assign directly, will change it

pytest_inmanta/plugin.py
```diff
@@ -226,10 +228,24 @@ def parse_repo(repo_str: str) -> object:
                 repo_info = module.ModuleRepoInfo(url=repo_str)
             if SUPPORTS_PROJECT_PIP_INDEX:
                 if repo_info.type == module.ModuleRepoType.package:
-                    LOGGER.warning(
-                        "Setting a package source through the --module-repo <index_url> with type `package` "
-                        "is now deprecated in favour of the --pip-index-urls <index_url> option.`"
+                    alternative_text: str = (
+                        "is now deprecated and will raise a warning during compilation."
+                        " Use the --pip-index-url <index_url> pytest option instead or set"
+                        "the %s environment variable to address these warnings. "
```
arnaudsjs, line 234:
```suggestion
                        " the %s environment variable to address these warnings. "
```

pytest_inmanta/plugin.py
```diff
@@ -226,10 +228,24 @@ def parse_repo(repo_str: str) -> object:
...
+                    if inm_mod_repo._value_set_using == ValueSetBy.ENV_VARIABLE:
+                        LOGGER.warning(
+                            "Setting a package source through the %s environment variable "
+                            + alternative_text,
+                            inm_mod_repo.environment_variable,
+                            pip_index_url.environment_variable,
+                        )
+                    elif inm_mod_repo._value_set_using == ValueSetBy.CLI:
+                        LOGGER.warning(
+                            "Setting a package source through the --module-repo <index_url> with type `package` "
```
arnaudsjs, line 245:
```suggestion
                            "Setting a package source through the --module-repo <index_url> cli option with type `package` "
```

arnaudsjs (changes requested):

sanderr (approved):

arnaudsjs (approved):

PR #422, opened 2023-10-11
Issue/unmanaged resources support

wouterdb:
# Description

1. I added support for unmanaged resources
2. I added a set of V2 functions

# TODO

- [x] tests
- [x] documentation

# Why V2

1. the current v1 is becoming very messy, because we learn as we go so it combines a lot of styles
2. refactoring the current code will break lots of other code
4. guidelines I want to follow on the V2
   - return wrapper objects (because these can be extended easily in the future without making a mess)
   - expose all intermediates (handler, context, ...) (turns out we need them at some point)
   - that have the relevant assertion functionality (make api easier to navigate by using OO)
 

# Self Check:

- [x] Changelog entry

### DIFF
--- a/examples/testmodule/model/_init.cf
+++ b/examples/testmodule/model/_init.cf
@@ -34,6 +34,16 @@ index ResourceBadIdAttribute(agent, name)
 implement ResourceBadIdAttribute using std::none
 
 entity Fail extends std::PurgeableResource:
+    """
+    A resource that fails unless value is `read` or purged is true
+
+    Logs a warning `Oh no!`
+
+    :attr agent: agent to deploy on
+    :attr name: value used as the resource id
+    :attr key: meaningless value
+    :attr value: meaningless value
+    """
     string name
     string agent
     string key
--- a/examples/testmodule/model/_init.cf
+++ b/examples/testmodule/model/_init.cf
@@ -32,3 +32,14 @@ end
 index ResourceBadIdAttribute(agent, name)
 
 implement ResourceBadIdAttribute using std::none
+
+entity Fail extends std::PurgeableResource:
+    string name
+    string agent
+    string key
+    string value
+end
+
+index Fail(agent, name)
+
+implement Fail using std::none
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -50,7 +50,7 @@ from inmanta.agent import handler
 from inmanta.agent import io as agent_io
 from inmanta.agent.handler import HandlerContext, ResourceHandler
 from inmanta.const import ResourceState
-from inmanta.data import DiscoveredResource, LogLine
+from inmanta.data import LogLine
 from inmanta.data.model import AttributeStateChange, ResourceIdStr
 from inmanta.execute.proxy import DynamicProxy
 from inmanta.export import Exporter, ResourceDict, cfg_env
@@ -462,10 +462,10 @@ class MockAgent(object):
 class MockClient(object):
     def __init__(self):
         self.facts = []
-        self.discovered_resources: List[DiscoveredResource] = []
+        self.discovered_resources: List[object] = []
 
     async def discovered_resource_create_batch(
-        self, tid, discovered_resources: collections.abc.Sequence[DiscoveredResource]
+        self, tid, discovered_resources: collections.abc.Sequence[object]
     ) -> Result:
         self.discovered_resources.extend(discovered_resources)
         return inmanta.protocol.common.Result(200)
@@ -1443,7 +1443,7 @@ class DeployResultV2:
         if change is not None:
             assert ctx._change == change
 
-    def discovered_resources(self) -> List[DiscoveredResource]:
+    def discovered_resources(self) -> List[object]:
         return self.handler._client.discovered_resources
 
 
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -1460,22 +1460,28 @@ class DeployResultV2:
     ):
         ctx = self.ctx
         if ctx.status != status:
-            print("Deploy did not result in correct status")
-            print("Requested changes: ", ctx._changes)
+            loglines = [
+                "Deploy did not result in correct status",
+                f"Requested changes: {ctx._changes}"
+                f"Outputting resource logs"
+            ]
+
             for log in ctx.logs:
-                print("Log: ", log._data["msg"])
-                print(
-                    "Kwargs: ",
-                    [
+                loglines.append(f"Log: {log._data['msg']}")
+                formattedkwargs = [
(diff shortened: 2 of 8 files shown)


pytest_inmanta/plugin.py
```diff
@@ -1358,6 +1376,76 @@ def finalize_all_handlers(self) -> None:
...
+        assert h is not None
+
+        ctx = handler.HandlerContext(resource, dry_run=dry_run)
+        h.execute(ctx, resource, dry_run)
+        self.finalize_context(ctx)
+        self.finalize_handler(h)
+
+        out = DeployResultV2(resource=resource, ctx=ctx, handler=h)
+        if status is not None:
+            out.assert_status(status)
```
wouterdb, line 1409: I added this to make it a bit similar to the V, 

(I made the mistake myself of not doing the assert and it took me some time)

pytest_inmanta/plugin.py
```diff
@@ -906,6 +923,7 @@ def get_handler(self, resource: Resource, run_as_root: bool) -> ResourceHandler:
             p.stat_file = lambda x: self.stat_blob(x)  # type: ignore
             p.upload_file = lambda x, y: self.add_blob(x, y)  # type: ignore
             p.run_sync = ioloop.IOLoop.current().run_sync  # type: ignore
+            p._client = MockClient()
```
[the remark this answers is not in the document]
edvgui: Do something similar to lsm mocked? And patch the existing client for the methods that we care about?

pytest_inmanta/plugin.py
```diff
@@ -1358,6 +1376,76 @@ def finalize_all_handlers(self) -> None:
...
+
+    def assert_status(
+        self,
+        status: const.ResourceState = const.ResourceState.deployed,
+        change: const.Change = None,
+    ):
+        ctx = self.ctx
+        if ctx.status != status:
+            print("Deploy did not result in correct status")
+            print("Requested changes: ", ctx._changes)
```
edvgui, line 1428: Why print? This will be lost if the test doesn't capture stdout, wich we usually not do

wouterdb: I copy-pasted this. But I agree that logging would probably be better

pytest_inmanta/plugin.py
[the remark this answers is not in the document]
wouterdb: On closer examination: the agent could never construct the client correctly, so nothing really changes

pytest_inmanta/plugin.py
wouterdb: otoh print is always shown when the tests fail, logging requires explicitly setting the log level.

pytest_inmanta/plugin.py
```diff
@@ -452,6 +457,20 @@ def __init__(self, uri):
...
         self.process = MockProcess()
         self._env_id = cfg_env.get()
+        self.sessionid = "mockid"
+        self.environment = self._env_id
+
+
+class MockClient(object):
+    def __init__(self):
+        self.facts = []
+        self.discovered_resources: List[object] = []
```
sanderr, line 467:
```suggestion
        self.discovered_resources: list[object] = []
```

pytest_inmanta/plugin.py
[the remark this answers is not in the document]
sanderr: > the mock agent could never construct the client correctly

Could you elaborate on that?

sanderr (changes requested):

pytest_inmanta/plugin.py
```diff
@@ -452,6 +455,20 @@ def __init__(self, uri):
         self.uri = uri
         self.process = MockProcess()
         self._env_id = cfg_env.get()
+        self.sessionid = "mockid"
+        self.environment = self._env_id
+
+
+class MockClient(object):
```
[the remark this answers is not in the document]
wouterdb: added

pytest_inmanta/plugin.py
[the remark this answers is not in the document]
wouterdb: the `get_client` method as it exists on `HandlerAPI`

```python
    def get_client(self) -> protocol.SessionClient:
        """
        Get the client instance that identifies itself with the agent session.

        :return: A client that is associated with the session of the agent that executes this handler.
        """
        if self._client is None:
            self._client = protocol.SessionClient("agent", self._agent.sessionid)
        return self._client
```

it requires 

`self._agent.sessionid`

but the `MockAgent` only got a `sessionid` field in this PR, so any previous attempt would have resulted in an attribute error

pytest_inmanta/plugin.py
wouterdb: opinion on this?

sanderr (approved):

examples/testmodule/model/_init.cf
```diff
@@ -32,3 +32,14 @@ end
 index ResourceBadIdAttribute(agent, name)
 
 implement ResourceBadIdAttribute using std::none
+
+entity Fail extends std::PurgeableResource:
+    string name
+    string agent
+    string key
+    string value
+end
```
edvgui, line 41: It might never get used by anyone else ever, but it would still be nice to document what it does and what the attributes are

pytest_inmanta/plugin.py
edvgui: I think it should be logs anyway

pytest_inmanta/plugin.py
```diff
@@ -1358,6 +1392,143 @@ def finalize_all_handlers(self) -> None:
...
+    @property
+    def logs(self) -> list[LogLine]:
+        return self.ctx.logs
+
+    def assert_has_logline(self, matches: str) -> LogLine:
+        """
+         Assert the handler logged a log line matching the pattern
+
+        :return: that logline
+        """
```
edvgui, line 1498: It could be made more clear that a regex is expected here.  Either by taking as input the pattern object, or documenting it clearly

pytest_inmanta/plugin.py
```diff
@@ -1358,6 +1392,143 @@ def finalize_all_handlers(self) -> None:
...
+    def assert_no_changes(self) -> None:
+        """Assert that the diff produced no changes"""
+        assert not self.changes
+
+    def assert_consistent_status(self):
+        """Make sure we report change and doing changes consistently"""
+        if self.ctx.status != const.ResourceState.deployed:
+            return
+
+        if bool(self.ctx.changes):
```
edvgui, line 1519: Why converting the dict to a bool? Does it changes something to simply doing this?
```python
if self.ctx.changes:
```

edvgui (changes requested):

edvgui (approved):

PR #424, opened 2023-10-16
issue/651-resource-propagates-childs-ignore

FloLey:
# Description

add a new resource entity that raises an IgnoreResourceException when buildings its id

linked to : https://github.com/inmanta/lsm/pull/657

closes https://github.com/inmanta/lsm/issues/651

# Self Check:

- [x] Changelog entry


pytest_inmanta/module/init.cf
```diff
@@ -54,3 +54,17 @@ end
 index IgnoreResource(name)
 
 implement IgnoreResource using std::none
+
+
+entity IgnoreResourceInId extends std::PurgeableResource:
+    """
+        This resource will raise an IgnoreResourceException when buildings its id.
```
Hugo-Inmanta, line 61:
```suggestion
        This resource will raise an IgnoreResourceException when building its id.
```

CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # v 2.9.0 (?)
 Changes in this release:
+- Adds a new resource entity that raises an IgnoreResourceException when buildings its id
```
Hugo-Inmanta, line 3: I'm not 100% sure whether a changelog entry is relevant here or not

arnaudsjs: I would keep it. I agree that it's unlikely that our customers will use this feature, but it's a new feature anyway.

arnaudsjs (changes requested):

arnaudsjs (changes requested):
It looks like you forgot to apply the rename consistency across the other parts of the PR.

pytest_inmanta/module/init.py
```diff
@@ -113,3 +113,19 @@ def get_desired_value(exporter, resource):
...
+    fields = ("name", "desired_value")
+
+    @staticmethod
+    def get_name(exporter, resource):
+        raise resources.IgnoreResourceException()
+
+
+@handler.provider("unittest::IgnoreResourceInIdAttr", name="ignore")
+class IgnoreResourceInIdAttrHandler(handler.CRUDHandler):
+    pass
```
arnaudsjs, line 131: Something you may be unaware of: You can put multiple `@handler.provider` annotations above a single class. So you could just as well have reused the `IgnoreResourceHandler` class for this.

arnaudsjs (approved):

pytest_inmanta/module/init.py
FloLey: Indeed, I didn't know. I'll change it

PR #429, opened 2023-11-07
Issue/iso7 pip

wouterdb:
# Description

Add new options for iso7, 

Companion to https://github.com/inmanta/inmanta-core/pull/6655

They should be merged in fairly close succession

# TODO:

- [ ] update backed in version constraint right before merge

# Self Check:

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@ Changes in this release:
 - Add support for unmanaged resources
 - Adds a new resource entity that raises an IgnoreResourceException when buildings its id
 - Add support for ISO7. 
-- Add `--pip-pre`, `--pip-use-system-config` and `--use-module-in-place`
+- Add `--pip-pre`, `--pip-use-system-config` and `--pip-index-url`
 
 # v 2.8.0 (2023-08-03)
 Changes in this release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Changes in this release:
 - Add `deploy_resouce_v2` endpoint
 - Add support for unmanaged resources
 - Adds a new resource entity that raises an IgnoreResourceException when buildings its id
+- Add support for ISO7. 
+- Add `--pip-pre`, `--pip-use-system-config` and `--use-module-in-place`
 
 # v 2.8.0 (2023-08-03)
 Changes in this release:
--- a/README.md
+++ b/README.md
@@ -219,14 +219,15 @@ The following options are available.
  * `--venv`: folder in which to place the virtual env for tests (will be shared by all tests), overrides `INMANTA_TEST_ENV`.
    This options depends on symlink support. This does not work on all windows versions. On windows 10 you need to run pytest in an
    admin shell. Using a fixed virtual environment can speed up running the tests.
+ * `--pip-index-url`: pip index to install dependencies from. Can be specified multiple times to add multiple indexes. When set, it will overwrite the system index-url even if `pip-use-system-config` is set. (overrides `INMANTA_PIP_INDEX_URL`, defaults to `[]`)
+ * `--pip-pre`, `--no-pip-pre` Allow installation of pre-release package by pip or not? (overrides `INMANTA_PIP_PRE`, defaults to  `--install-mode != release` )
+ * `--pip-use-system-config` Allow pytest-inmanta to use the system pip config or not? (overrides `INMANTA_PIP_USE_SYSTEM_CONFIG`, defaults to `False`)
  * `--use-module-in-place`: makes inmanta add the parent directory of your module directory to it's directory path, instead of copying your
     module to a temporary libs directory. It allows testing the current module against specific versions of dependent modules. 
     Using this option can speed up the tests, because the module dependencies are not downloaded multiple times.
- * `--module-repo`: location to download modules from, overrides `INMANTA_MODULE_REPO`. The default value is the inmanta github organisation.
-    For versions of inmanta-core that support v2 modules, the repo accepts the format "[\<type\>:]\<url\>" with "type" the repository type as
-    defined in the project config documentation. If type is omitted, git is assumed.
+ * `--module-repo`: location to download v1 modules from, overrides `INMANTA_MODULE_REPO`. The default value is the inmanta github organisation.
     Multiple repos can be passed by space-separating them or by passing the parameter multiple times.
- * `--install-mode`: install mode to use for modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`.
+ * `--install-mode`: install mode to use for v1 modules downloaded during this test, overrides `INMANTA_INSTALL_MODE`.
  * `--no-load-plugins`: Don't load plugins in the Project class. Overrides `INMANTA_NO_LOAD_PLUGINS`. 
  When not using this option during the testing of plugins with the `project.get_plugin_function` method, 
  it's possible that the module's `plugin/__init__.py` is loaded multiple times, 
--- a/README.md
+++ b/README.md
@@ -219,8 +219,8 @@ The following options are available.
  * `--venv`: folder in which to place the virtual env for tests (will be shared by all tests), overrides `INMANTA_TEST_ENV`.
    This options depends on symlink support. This does not work on all windows versions. On windows 10 you need to run pytest in an
    admin shell. Using a fixed virtual environment can speed up running the tests.
- * `--pip-index-url`: pip index to install dependencies from. Can be specified multiple times to add multiple indexes. When set, it will overwrite the system index-url even if `pip-use-system-config` is set. (overrides `INMANTA_PIP_INDEX_URL`, defaults to `[]`)
- * `--pip-pre`, `--no-pip-pre` Allow installation of pre-release package by pip or not? (overrides `INMANTA_PIP_PRE`, defaults to  `--install-mode != release` )
+ * `--pip-index-url`: pip index to install dependencies from. Can be specified multiple times to add multiple indexes. When set, it will overwrite the system index-url even if `pip-use-system-config` is set. (overrides `PIP_INDEX_URL`, defaults to `[]`)
+ * `--pip-pre`, `--no-pip-pre` Allow installation of pre-release package by pip or not? (overrides `PIP_PRE`, defaults to  `--install-mode != release` )
  * `--pip-use-system-config` Allow pytest-inmanta to use the system pip config or not? (overrides `INMANTA_PIP_USE_SYSTEM_CONFIG`, defaults to `False`)
  * `--use-module-in-place`: makes inmanta add the parent directory of your module directory to it's directory path, instead of copying your
     module to a temporary libs directory. It allows testing the current module against specific versions of dependent modules. 
--- a/pytest_inmanta/core.py
+++ b/pytest_inmanta/core.py
@@ -38,7 +38,10 @@ except DistributionNotFound:
 # Setting a project-wide pip index is only supported for iso7+
 SUPPORTS_PROJECT_PIP_INDEX_ISO7: bool = (
     CORE_VERSION is not None
-    and CORE_VERSION >= version.Version("10.0.0.dev20231111")  # TODO: set correctly
+    and CORE_VERSION
+    >= version.Version(
+        "11.0.0.dev20231110000000"
+    )  # TODO: set correctly, now set to make backward compat tests succeed
 )
 
 
--- a/pytest_inmanta/core.py
+++ b/pytest_inmanta/core.py
@@ -40,8 +40,8 @@ SUPPORTS_PROJECT_PIP_INDEX: bool = (
     CORE_VERSION is not None
     and CORE_VERSION
     >= version.Version(
-        "11.0.0.dev20231110000000"
-    )  # TODO: set correctly, now set to make backward compat tests succeed
+        "11.0.0.dev"
+    )
 )
 
 
--- a/pytest_inmanta/core.py
+++ b/pytest_inmanta/core.py
@@ -40,7 +40,7 @@ SUPPORTS_PROJECT_PIP_INDEX: bool = (
     CORE_VERSION is not None
     and CORE_VERSION
     >= version.Version(
-        "10.0.0.dev"
+        "11.0.0.dev20231110000000"
     )  # TODO: set correctly, now set to make backward compat tests succeed
 )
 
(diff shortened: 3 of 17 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changes in this release:
 - Add `deploy_resouce_v2` endpoint
 - Add support for unmanaged resources
 - Adds a new resource entity that raises an IgnoreResourceException when buildings its id
+- Add support for ISO7. 
+- Add `--pip-pre`, `--pip-use-system-config` and `--use-module-in-place`
```
edvgui, line 7:
```suggestion
- Add `--pip-pre`, `--pip-use-system-config` and `--pip-index-url`
```
I guess?

pytest_inmanta/test_parameter/boolean_parameter.py
```diff
@@ -51,7 +52,7 @@ def __init__(
         environment_variable: str,
         usage: str,
         *,
-        default=False,
+        default: Optional[Union[bool, DynamicDefault[bool]]] = False,
```
edvgui, line 55: Is this some leftover from before creating the new type?

edvgui (changes requested):
I did a very quick review, I need to dig in a bit more to understand what is actually going on

pytest_inmanta/test_parameter/boolean_parameter.py
wouterdb: No, why?

wouterdb: No, why?

Hugo-Inmanta (changes requested):

Hugo-Inmanta (approved):

pytest_inmanta/test_parameter/boolean_parameter.py
edvgui: Nevermind

edvgui, line 55: Can you also update the signature of the `__init__` method for the other classes?

wouterdb: done

edvgui (approved):

edvgui (changes requested):

pytest_inmanta/parameters.py
```diff
@@ -147,7 +181,7 @@ def resolve(self, config: "Config") -> bool:
         if os.getenv(self.environment_variable):
             return True
 
-        return False
```
wouterdb, line 150: This cause this warning to always be produced.

pytest_inmanta/parameters.py
```diff
@@ -142,6 +143,7 @@ def get_help(self) -> str:
 pip_index_url = ListTestParameter(
     argument="--pip-index-url",
     environment_variable="PIP_INDEX_URL",
+    legacy_environment_variable="INMANTA_PIP_INDEX_URL",
```
edvgui, line 146: Why? This never was an environment variable we accept, was it?

wouterdb: it was (see the diff)

edvgui: Oh, indeed, sorry

edvgui (approved):

pytest_inmanta/plugin.py
```diff
@@ -293,19 +301,52 @@ def project_metadata(request: pytest.FixtureRequest) -> module.ProjectMetadata:
...
+            pre=pip_pre,
+        )
+
+        if not pip_config.has_source():
+            LOGGER.warning(PIP_NO_SOURCE_WARNING)
+
+        return module.ProjectMetadata(
+            name="testcase",
+            description="Project for testcase",
+            repo=repos,
```
sanderr, line 342: Passing on the package type repos to core will trigger a warning. We used to do this because core would still use it for backwards compatibility, but that is no longer the case. So I think we should only pass on the v1 repos, and update our warning message in this case (it states that the compiler will raise a warning).

pytest_inmanta/test_parameter/optional_boolean_parameter.py
```diff
@@ -0,0 +1,57 @@
...
+    Contact: code@inmanta.com
+"""
+import argparse
+from typing import Optional, Type, Union
+
+from pytest_inmanta.test_parameter.boolean_parameter import BooleanTestParameter
+from pytest_inmanta.test_parameter.parameter import DynamicDefault
+
+
+class OptionalBooleanTestParameter(BooleanTestParameter):
```
sanderr, line 25: I have a tiny preference to put this in the same module as the normal `BooleanTestParameter`. I think if I were searching for this class, that's where I'd look (grep aside).

sanderr (changes requested):
I added some comments regarding the remote orchestrator. I now realize this is pytest-inmanta and the remote orchestrator is a pytest-inmanta-lsm concept. I'm not sure how best to handle this? Should we just put this in the pytest-inmanta-lsm readme?

README.md
```diff
@@ -219,14 +219,15 @@ The following options are available.
  * `--venv`: folder in which to place the virtual env for tests (will be shared by all tests), overrides `INMANTA_TEST_ENV`.
    This options depends on symlink support. This does not work on all windows versions. On windows 10 you need to run pytest in an
    admin shell. Using a fixed virtual environment can speed up running the tests.
+ * `--pip-index-url`: pip index to install dependencies from. Can be specified multiple times to add multiple indexes. When set, it will overwrite the system index-url even if `pip-use-system-config` is set. (overrides `PIP_INDEX_URL`, defaults to `[]`)
+ * `--pip-pre`, `--no-pip-pre` Allow installation of pre-release package by pip or not? (overrides `PIP_PRE`, defaults to  `--install-mode != release` )
```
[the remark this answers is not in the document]
wouterdb: Unsure, but this is more-or-less backward compatible

pytest_inmanta/core.py
```diff
@@ -36,6 +36,15 @@
     CORE_VERSION = None
 
 # Setting a project-wide pip index is only supported for iso7+
+SUPPORTS_PROJECT_PIP_INDEX_ISO7: bool = (
+    CORE_VERSION is not None
+    and CORE_VERSION
+    >= version.Version(
+        "11.0.0.dev20231110000000"
+    )  # TODO: set correctly, now set to make backward compat tests succeed
```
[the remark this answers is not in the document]
wouterdb: Yes, needs to be cleaned up after merge of core

README.md
```diff
@@ -219,14 +219,15 @@ The following options are available.
...
+ * `--pip-use-system-config` Allow pytest-inmanta to use the system pip config or not? (overrides `INMANTA_PIP_USE_SYSTEM_CONFIG`, defaults to `False`)
```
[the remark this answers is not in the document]
wouterdb: There is no such thing here,...

pytest_inmanta/parameters.py
```diff
@@ -128,6 +121,47 @@
...
+        "When set, it will overwrite the system index-url even if pip-use-system-config is set."
+    ),
+    group=param_group,
+    default=[],
+)
+
+pip_use_system_config = BooleanTestParameter(
+    argument="--pip-use-system-config",
+    environment_variable="INMANTA_PIP_USE_SYSTEM_CONFIG",
+    usage=("Allow pytest-inmanta to use the system pip config or not?"),
```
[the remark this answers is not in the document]
wouterdb: There is no such thing here,...

pytest_inmanta/core.py
```diff
@@ -36,6 +36,15 @@
     CORE_VERSION = None
 
 # Setting a project-wide pip index is only supported for iso7+
+SUPPORTS_PROJECT_PIP_INDEX_ISO7: bool = (
```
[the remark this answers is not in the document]
wouterdb: What would you suggest?

pytest_inmanta/parameters.py
```diff
@@ -128,6 +121,47 @@
...
+        return install_mode != InstallMode.release
+
+    def get_help(self) -> str:
+        return " --install-mode != release "
+
+
+pip_pre = OptionalBooleanTestParameter(
+    argument="--pip-pre",
+    environment_variable="PIP_PRE",
+    usage=("Allow installation of pre-release package by pip or not?"),
```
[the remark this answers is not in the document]
wouterdb: well, due to lack of isolation, it may have worked before. Not sure

pytest_inmanta/test_parameter/optional_boolean_parameter.py
wouterdb: They are currently one-file-per-class, so I kept to that (even if I would put all of then in one file)

README.md
```diff
@@ -219,14 +219,15 @@ The following options are available.
  * `--venv`: folder in which to place the virtual env for tests (will be shared by all tests), overrides `INMANTA_TEST_ENV`.
    This options depends on symlink support. This does not work on all windows versions. On windows 10 you need to run pytest in an
    admin shell. Using a fixed virtual environment can speed up running the tests.
+ * `--pip-index-url`: pip index to install dependencies from. Can be specified multiple times to add multiple indexes. When set, it will overwrite the system index-url even if `pip-use-system-config` is set. (overrides `PIP_INDEX_URL`, defaults to `[]`)
```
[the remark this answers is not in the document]
wouterdb: crucial difference is that we allow it to be set multiple times, where all but the first become the `PIP_EXTRA_INDEX_URL` 

I don't know how to document this without making it more confusing.

pytest_inmanta/parameters.py
```diff
@@ -128,6 +121,47 @@
 )
 
 
+class PipPreFallbackToReleaseMode(DynamicDefault[bool]):
```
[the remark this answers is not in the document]
wouterdb: updated

pytest_inmanta/plugin.py
wouterdb: I don't see why we need to update this?

Here, it still works in a backward compatible way and the compile will still raise a warning ?

pytest_inmanta/test_parameter/optional_boolean_parameter.py
wouterdb: If you want, I'll move it.

pytest_inmanta/plugin.py
sanderr: The compiler ignores it now, doesn't it?

sanderr: I mean, it still raises a warning, but it is otherwise ignored.

pytest_inmanta/parameters.py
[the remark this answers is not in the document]
sanderr: But the cmdline option will not have any effect on e.g. iso6

pytest_inmanta/core.py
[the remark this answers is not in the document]
sanderr: I don't feel too strongly about this, but perhaps make this one `SUPPORTS_PROJECT_PIP_INDEX` and the other one `SUPPORTS_LEGACY_PROJECT_PIP_INDEX`? The old one really only exists for one or two versions of OSS and that's it.

README.md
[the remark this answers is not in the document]
sanderr: Indeed, I added a comment about that [here](https://github.com/inmanta/pytest-inmanta/pull/429#pullrequestreview-1722035907). Wdyt? Should we just add something to the pytest-inmanta-lsm readme?

pytest_inmanta/core.py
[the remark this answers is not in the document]
wouterdb: done

pytest_inmanta/parameters.py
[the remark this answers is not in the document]
wouterdb: no, indeed.

pytest_inmanta/plugin.py
wouterdb: Yes, but we convert them into pip config here as well.

pytest_inmanta/parameters.py
```diff
@@ -135,9 +135,11 @@ def get_help(self) -> str:
 pip_pre = OptionalBooleanTestParameter(
     argument="--pip-pre",
     environment_variable="PIP_PRE",
-    usage=("Allow installation of pre-release package by pip or not?"),
+    usage=(
+        "Allow installation of pre-release package by pip or not? (only for inmanta-core>=11, >ISO7)"
```
sanderr, line 139:
```suggestion
        "Allow installation of pre-release package by pip or not? (only for inmanta-core>=11, >=ISO7)"
```

pytest_inmanta/parameters.py
```diff
@@ -156,7 +158,9 @@ def get_help(self) -> str:
 pip_use_system_config = BooleanTestParameter(
     argument="--pip-use-system-config",
     environment_variable="INMANTA_PIP_USE_SYSTEM_CONFIG",
-    usage=("Allow pytest-inmanta to use the system pip config or not?"),
+    usage=(
+        "Allow pytest-inmanta to use the system pip config or not? (only for inmanta-core>=11, >ISO7)"
```
sanderr, line 162:
```suggestion
        "Allow pytest-inmanta to use the system pip config or not? (only for inmanta-core>=11, >=ISO7)"
```

pytest_inmanta/plugin.py
sanderr: Yes, but we still trigger the warning from core, which is not tailored to pytest-inmanta usage.

README.md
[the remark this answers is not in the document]
sanderr: Fair enough. If `use-system-config` means you are responsible for configuring all components correctly, then I guess that implies that you need to manually configure the remote orchestrator.

README.md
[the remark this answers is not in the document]
sanderr: Yes, for the write direction. But we don't read `PIP_EXTRA_INDEX_URL`. We completely override it with what is supplied here.

README.md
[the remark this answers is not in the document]
sanderr: I would like to get rid of it, if we can. It's a big if. I think most Jenkinsfiles set `PIP_PRE` anyway, so it could be worth a shot? But perhaps merge it in a second stage (meaning we shouldn't release pytest-inmanta yet) to isolate fallout?

pytest_inmanta/plugin.py
wouterdb: done

sanderr (approved):

Issue #437, opened 2023-12-07
Resources are not fully serialized on compile


edvgui:
The resource serialization logic only runs the exporter but doesn't simulate the api boundary serialization.  

This is an issue because:
1. we can end up with some proxy objects in the handler code when running tests.
2. we don't test that the serialized resource's attributes can actually be serialized into json and sent to the orchestrator

The proposed fix is to add something like this at the end of the compile/export code:

```python
# Fix the resource serialization (fully serialize them)
new_resources = {}
for resource_id, resource in project.resources.items():
    new_resource = inmanta.resources.Resource.deserialize(
        json.loads(inmanta.protocol.common.json_encode(resource.serialize()))
    )
    new_resource.model = resource.model
    new_resources[resource_id] = new_resource

project.resources = new_resources
```

PR #456, opened 2024-03-21
Fix resource unserialized when deploy_all is used

hlloreda:
# Description

When `deploy_all` was used, the resource was not serialized correctly. This led to the presence of proxy instances in the different fields of the resource, which could cause problems when interacting with an API.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # v 2.10.0 (?)
 Changes in this release:
 - Add support for custom `agent_map` enabling testing or remote io scenarios.
+- Make sure that the resource is serialized when `deploy_all` is used
 
 # v 2.9.0 (2023-11-29)
 Changes in this release:
--- a/Makefile
+++ b/Makefile
@@ -5,9 +5,8 @@ black = black pytest_inmanta tests examples
 
 .PHONY: install
 install:
-	pip install -U setuptools pip
-	pip install -U -r requirements.txt -r requirements.dev.txt
-	pip install -e .
+	pip install -U --upgrade-strategy=eager pip setuptools wheel
+	pip install -U --upgrade-strategy=eager -e . -r requirements.txt -r requirements.dev.txt
 
 .PHONY: format
 format:
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -741,6 +741,12 @@ def get_resources_matching(
         yield resource
 
 
+def check_serialization(resource: Resource) -> Resource:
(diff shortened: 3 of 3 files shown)


pytest_inmanta/plugin.py
```diff
@@ -1444,11 +1446,6 @@ def unittest_resource_set(
         """
         DATA[name].update(kwargs)
 
-    def check_serialization(self, resource: Resource) -> Resource:
-        """Check if the resource is serializable"""
-        serialized = json.loads(protocol.json_encode(resource.serialize()))
-        return Resource.deserialize(serialized)
-
```
edvgui, line 1451: This is a breaking change, why are you removing this?

Makefile
```diff
@@ -5,9 +5,8 @@ black = black pytest_inmanta tests examples
 
 .PHONY: install
 install:
-	pip install -U setuptools pip
-	pip install -U -r requirements.txt -r requirements.dev.txt
-	pip install -e .
+	pip install -U --upgrade-strategy=eager pip setuptools wheel
+	pip install -U --upgrade-strategy=eager -e . -r requirements.txt -r requirements.dev.txt
```
edvgui, line 9:
```suggestion
	pip install -U --upgrade-strategy=eager -e . -c requirements.txt -r requirements.dev.txt
```
Looks like the file is a lock file here

edvgui (changes requested):
Did you have a look at this ticket first? https://github.com/inmanta/pytest-inmanta/issues/437
Maybe we want to fix this during compile and not during deploy

pytest_inmanta/plugin.py
```diff
@@ -1320,9 +1321,18 @@ def enumerate_model(model: str):
                     DeprecationWarning,
                 )
 
+        # Fix the resource serialization (fully serialize them)
+        new_resources = {}
+        for resource_id, resource in self.resources.items():
```
edvgui, line 1326: That will not work, you are accessing the resources before setting them (you are actually reading the resources from the previous compile)

CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 # v 2.10.0 (?)
 Changes in this release:
 - Add support for custom `agent_map` enabling testing or remote io scenarios.
+- Make sure that the resource is serialized when `deploy_all` is used
```
edvgui, line 4:
```suggestion
- Fix resource serialization (#437)
```

edvgui (changes requested):

edvgui (approved):

Issue #468, opened 2024-05-30
Make a helper method on the project fixture to find one specific resource


wouterdb:
- the current method filters on the model type, not on the resource type, which causes subtle issues
- is doesn't assert there is only one

PR #473, opened 2024-07-03
Add stricter get_resource method

hlloreda:
# Description

- Added `Project.get_one_resource` method to have a stricter filtering
- Added strict mode on `Project.get_resource` method (if set to `True`, will internally call get_one_resource)

closes https://github.com/inmanta/pytest-inmanta/issues/468

# Self Check:

- [x] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,8 +3,11 @@ Changes in this release:
 - Add support for custom `agent_map` enabling testing or remote io scenarios.
 - Fix resource serialization (#437)
 - Fix new agent api in inmanta.agent.executor.AgentInstance (ISO8)
-- Add `get_one_resource` method to the `project` fixture. This method is the upgraded version of `get_resource`. This new method should be used instead of the old one and any user is encouraged to migrate to the new one. Please refer to the documentation of the fixture for further information.
-- Add strict mode in `get_resource` method to  the `project` fixture (by default set to `False`). If strict mode is set to `True`, the new `get_one_resource` method will be called. 
+- Add `get_one_resource` method to the `project` fixture and `Result` object.
+- Add strict mode in `get_resource` method to  the `project` fixture and `Result` object (by default set to `False`). 
+
+## Updating
+- All uses of the `get_resource` method should be replaced with the `get_one_resource` method. It has improved type matching and raises an exception when multiple matches are present. 
 
 # v 2.9.0 (2023-11-29)
 Changes in this release:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ Changes in this release:
 - Add support for custom `agent_map` enabling testing or remote io scenarios.
 - Fix resource serialization (#437)
 - Fix new agent api in inmanta.agent.executor.AgentInstance (ISO8)
+- Add `get_one_resource` method to the `project` fixture. This method is the upgraded version of `get_resource`. This new method should be used instead of the old one and any user is encouraged to migrate to the new one. Please refer to the documentation of the fixture for further information.
+- Add strict mode in `get_resource` method to  the `project` fixture (by default set to False). If strict mode is set to True, the new `get_one_resource` method will be called. 
 
 # v 2.9.0 (2023-11-29)
 Changes in this release:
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -3,6 +3,8 @@ Changes in this release:
 - Add support for custom `agent_map` enabling testing or remote io scenarios.
 - Fix resource serialization (#437)
 - Fix new agent api in inmanta.agent.executor.AgentInstance (ISO8)
+- Add `get_one_resource` method to the `project` fixture. This method is the upgraded version of `get_resource`. This new method should be used instead of the old one and any user is encouraged to migrate to the new one. Please refer to the documentation of the fixture for further information.
+- Add strict mode in `get_resource` method to  the `project` fixture (by default set to `False`). If strict mode is set to `True`, the new `get_one_resource` method will be called.
```
wouterdb, line 7:
```suggestion
- Add `get_one_resource` method to the `project` fixture and `Result` object.
- Add strict mode in `get_resource` method to  the `project` fixture and `Result` object (by default set to `False`). 


## Updating
- All uses of the `get_resource` method should be replaced with the `get_one_resource` method. It has improved type matching and raises an exception when multiple matches are present. 
```

this make it visually apparent that there is something to do

examples/testhandler/tests/test_get_resource_subtle_case.py
```diff
@@ -0,0 +1,70 @@
...
+        """
+    import unittest
+
+    unittest::Resource(name="res", desired_value="x")
+    unittest::ResourceA(name="fakeres", desired_value="fakex")
+    """
+    )
+
+    resource = project.get_resource("unittest::Resource")
+    wrong_resource_a = project.get_resource("unittest::ResourceA")
```
wouterdb, line 42: I expect this to be flaky: there are two object that can be returned and no way to know which on is the correct one.

I would not do ```@resources.resource("unittest::ResourceA", id_attribute="name", agent="agent")```

and then test

```
resource = project.get_resource("unittest::Resource", name="fakeres" )    # will not find anything in normal mode, but will in strict mode
resource = project.get_resource("unittest::ResourceA", name="fakeres" )  # will not find anything in strict mode, but will in normal mode
 project.get_resource("unittest::Resource") # will fail in strict mode, succeed in normal mode
```

pytest_inmanta/plugin.py
```diff
@@ -735,7 +747,10 @@ def apply_filter(resource: Resource) -> bool:
         return True
 
     for resource in resources:
-        if not resource.is_type(resource_type):
+        match_entity_type = resource.id.entity_type == resource_type
+        if not resource.is_type(resource_type) or (
```
wouterdb, line 751: this is not correct as it still uses the older filter as well

pytest_inmanta/plugin.py
```diff
@@ -763,13 +779,52 @@ def get_resource(
...
+    list_resources = []
+    for resource in resources:
+        resource = check_serialization(resource)
+        list_resources.append(resource)
+
+    # If we don't find anything matching these criteria, we should return `None`
+    if len(list_resources) == 0:
+        list_resources.append(None)
+
+    assert len(list_resources) == 1, "Only one resource should be found!"
```
wouterdb, line 824: I would provide context to help debug this:

```suggestion
    assert len(list_resources) == 1, "The filter should only match one resource, but it matches: [{','.join(str(resource.id) for resource in list_resource)}]"
```

pytest_inmanta/plugin.py
```diff
@@ -820,6 +875,23 @@ def get_context_for(
...
+
+        :param resource_type: The exact type used in the model (no super types)
+        :param strict_mode: If we need to make additional assertion when retrieving the resource:
+            - stricter filtering: matching the entity type of the resource
+            - assert only one instance
+        """
+        if strict_mode:
+            return self.get_one_resource(resource_type, **filter_args)
+        else:
+            return get_resource(self.results.keys(), resource_type, **filter_args)
```
wouterdb, line 892: Why not pass the strict mode in here ?  (detail)

wouterdb (changes requested):

pytest_inmanta/module/init.py
```diff
@@ -30,6 +30,7 @@ class Resource(resources.PurgeableResource):
     fields = ("name", "desired_value", "skip", "fail", "fail_deploy", "wrong_diff")
 
 
+@handler.provider("unittest::ResourceA", name="test")
```
wouterdb, line 33: this can go

wouterdb (approved):

PR #481, opened 2024-08-19
Remove IO support

bartv:
# Description

* Short description here *

closes *Add ticket reference here*

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
 # v 2.11.0 (?)
 Changes in this release:
+- Remove io support to be compatible with ISO8
 
 # v 2.10.0 (2024-07-05)
 Changes in this release:
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -45,8 +45,9 @@ from tornado import ioloop
 
 import inmanta.ast
 from inmanta import compiler, config, const, module, plugins, protocol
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -1,5 +1,6 @@
 # v 2.11.0 (?)
```
wouterdb, line 1: I wonder if we need to make this a 3.0.0 if we drop IO ? 

We could, if we want to, put the import `from inmanta.agent import io as agent_io` in the method body and keep support for iso7 ?

wouterdb (approved):

CHANGELOG.md
bartv: only std and exec used it and those modules have been ported and updated.

Depends if we want to keep supporting the iso7 branch of std with the latest pytest-inmanta


I agree with the major bump.

bartv: Given that the std resources are only used by us internally, I am inclined to just handle it in the std iso7 branch instead of poluting pytest-inmanta even more

bartv: For the version: isn't that the release tool that bumps this to 3.0.0?

wouterdb: good question, I'm not sure

wouterdb (approved):

PR #492, opened 2024-10-10
improve relative path for PathTestParameter

FelixGaudin:
# Description

* Short description here *

closes *Add ticket reference here*

# Self Check:

- [ ] ~~Attached issue to pull request~~
- [x] Changelog entry
- [ ] ~~Sufficient test cases (reproduces the bug/tests the requested feature)~~
- [ ] ~~Correct, in line with design~~
- [ ] ~~End user documentation is included or an issue is created for end-user documentation (add ref to issue here: )~~

### DIFF
--- a/pytest_inmanta/test_parameter/path_parameter.py
+++ b/pytest_inmanta/test_parameter/path_parameter.py
@@ -25,6 +25,9 @@ from pytest_inmanta.test_parameter.parameter import DynamicDefault, TestParamete
 
 
 def abspath(path: str, *, source: str) -> str:
+    """
+    Compute absolute path from relative or absolute path
+    """
     return path if os.path.isabs(path) else os.path.abspath(os.path.join(source, path))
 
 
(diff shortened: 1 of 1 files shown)


pytest_inmanta/test_parameter/path_parameter.py
```diff
@@ -25,18 +24,11 @@
 from pytest_inmanta.test_parameter.parameter import DynamicDefault, TestParameter
 
 
-@contextlib.contextmanager
-def working_dir(path: str):
+def abspath(path: str, *, source: str) -> str:
```
sanderr, line 27: I don't see particular value in having a dedicated method for this, but I'm not against. it. If we do, the `source` should have a `:param:` in the docstring, because it's not trivial.

sanderr (changes requested):

sanderr (approved):

PR #497, opened 2024-11-05
Issue/no pydantic warning

edvgui:
# Description

Some fixtures were still using pydantic v1 by default, which raises warnings in every ci.  This addresses it.  I don't know if we still support pydantic v1 for any version of the product, I kept it here by default, but maybe we can drop it.

# Self Check:

- [x] Changelog entry

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -258,7 +258,12 @@ def get_project_repos(repo_options: typing.Sequence[str]) -> typing.Sequence[obj
                             + alternative_text,
                             parameters.pip_index_url.environment_variable,
                         )
-            return json.loads(repo_info.json())
+
+            try:
+                return repo_info.model_dump(mode="json")
+            except AttributeError:
+                # Backward compatibility with pydantic v1
+                return json.loads(repo_info.json())
 
     return [parse_repo(repo) for repo in repo_options]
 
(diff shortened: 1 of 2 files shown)


pytest_inmanta/plugin.py
```diff
@@ -258,7 +258,12 @@ def parse_repo(repo_str: str) -> object:
                             + alternative_text,
                             parameters.pip_index_url.environment_variable,
                         )
-            return json.loads(repo_info.json())
+
+            try:
+                return repo_info.model_dump(mode="json")
+            except AttributeError:
+                # Backward compatibility with pydantic v1
+                return json.loads(repo_info.json())
```
sanderr, line 266: > I don't know if we still support pydantic v1 for any version of the product

We don't

pytest_inmanta/plugin.py
```diff
@@ -258,7 +258,12 @@ def parse_repo(repo_str: str) -> object:
                             + alternative_text,
                             parameters.pip_index_url.environment_variable,
                         )
-            return json.loads(repo_info.json())
+
+            try:
+                return repo_info.model_dump(mode="json")
```
sanderr, line 263: I didn't know about the json mode, that's neat!

sanderr (changes requested):

pytest_inmanta/plugin.py
edvgui: Perfect, I dropped it

sanderr (approved):

PR #524, opened 2025-04-14
Issue/clear logs before compile

edvgui:
# Description

- Clear pytest's in-memory logs before each compile to avoid heavy memory leaks

# Self Check:

- [x] Changelog entry


tests/test_basic_example_v2.py
```diff
@@ -95,7 +95,7 @@ def test_basic_example(
                 "The module being tested is not installed in editable mode."
                 " As a result the tests will not pick up any changes to the local source files."
                 " To install it in editable mode, run `pip install -e .`."
-                in caplog.messages
+                in caplog.text
```
edvgui, line 98: The manual `clear` done before each compiles removes the LogRecord object from the list, but preserves the StringIO so the text still contains it

sanderr (approved):

PR #541, opened 2025-07-25
Fix incompatibility inmanta-core: push environment settings

arnaudsjs:
# Description

Fix incompatibility with versions of inmanta-core that have support to push environment settings to the server.

Related to https://github.com/inmanta/inmanta-core/pull/9355

# Self Check:

- [x] Changelog entry
- [ ] ~~End user documentation is included or an issue is created for end-user documentation~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 Changes in this release:
 - Add support for skip in reference
 - Fix memory leak caused by pytest LogCaptureHandler.  Address it by clearing the old in-memory logs before each compile.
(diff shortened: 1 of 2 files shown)


Hugo-Inmanta (approved):

arnaudsjs:
@Hugo-Inmanta I requested re-review because I realized that my fix was not entirely safe. Resources are only exported when there are resources present in the model. The call to export environment settings happens always, so there are situation where pytest-inmanta would not go to the server with `export=True`, while it would in the new code. The latest fix makes sure we only export environment settings when explicitly requested.

PR #549, opened 2025-09-15
test unloading

### LINKED ISSUE #526: Subsequent exports of model with different imports fail
# Problem

Given a testcase that uses lsm and exports to the server. 
If the next testcase doesn't use lsm and export to the server, we get 

```python
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
[...]/pytest_inmanta/plugin.py:1408: in compile
    version, resources = exporter.run(types, scopes, no_commit=not export)
[...]/inmanta/export.py:442: in run
    self._version = self.commit_resources(
[...]/inmanta/export.py:560: in commit_resources
    self.deploy_code(conn, tid, version)
[...]/inmanta/export.py:516: in deploy_code
    code_manager.register_code(type_name, resource_definition)
[...]/inmanta/loader.py:158: in register_code
    all_plugin_files: list[SourceInfo] = self._get_source_info_for_module(module_name)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <inmanta.loader.CodeManager object at 0x7fdfec367e30>, module_name = 'lsm'

    def _get_source_info_for_module(self, module_name: str) -> list[SourceInfo]:
        if module_name in self.__module_to_source_info:
            return self.__module_to_source_info[module_name]
    
        sources = [
>           SourceInfo(path, module_name) for path, module_name in module.Project.get().modules[module_name].get_plugin_files()
        ]
E       KeyError: 'lsm'

[...]/inmanta/loader.py:167: KeyError
-------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- Captured log setup --------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
INFO     inmanta.env:env.py:727 Initializing virtual environment at /tmp/tmpk3uv3cb5/.env

```

# Work around

```
@pytest.fixture()
def project(project: Project) -> Project:
    yield project

    # Clean everything that can persists between tests
    to_remove = []
    for resource_name in resource._resources:
        if not resource_name.startswith("std::"):
            to_remove.append(resource_name)

    for resource_name in to_remove:
        del resource._resources[resource_name]

    Commander.reset()
```

wouterdb:
# Description
1. this fixes the problem
2. testcase is simple, but requires `pytest-inmanta-lsm`. Where do I put it?  I'm in favor of putting it there. 

```
def test_reload(project, remote_orchestrator):
    project.compile("import lsm", export=True)
    project.compile('std::print("test")', export=True)
```

3. should I move this into core, if so, which versions? 

closes #526 

# Self Check:

- [ ] Changelog entry

### DIFF
--- a/pytest_inmanta/plugin.py
+++ b/pytest_inmanta/plugin.py
@@ -627,6 +627,7 @@ class ProjectLoader:
     def _unregister_old(cls, project: module.Project):
         """
         Resources, handlers and references are not unloaded or re-registered
+        by default in-between runs.
 
         When not importing a module on a second run we need to de-register them
         to prevent the exporter from failing when it tries to export them.


pytest_inmanta/plugin.py
```diff
@@ -612,6 +623,41 @@ def _register_plugins(cls, project: module.Project) -> None:
             ):
                 plugins.PluginMeta.add_function(plugin)
 
+    @classmethod
+    def _unregister_old(cls, project: module.Project):
+        """
+        Resources, handlers and references are not unloaded or re-registered
+
```
Hugo-Inmanta, line 630:
```suggestion
        Resources, handlers and references are not unloaded or re-registered
        by default in-between runs.

```

pytest_inmanta/plugin.py
```diff
@@ -612,6 +623,41 @@ def _register_plugins(cls, project: module.Project) -> None:
             ):
                 plugins.PluginMeta.add_function(plugin)
 
+    @classmethod
+    def _unregister_old(cls, project: module.Project):
```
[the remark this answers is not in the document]
Hugo-Inmanta: > It would probably make sense to move this into core.

I think I agree, I assume the signature would be slightly different e.g. transform current `cls._all_previous_modules` into an argument and return it afterwards.
If we do, we should mention in the docstring that this method is purely intended during specific tests scenarii where resetting the project in-between runs would be too cumbersome (Is this correct ? I'm not 100% sure ?)

> Question is: which versions of core?

Versions with active development I would assume ? 8 and 9 ? Do you foresee any difficulties with iso8 ? Or were you asking about whether this should also go into iso7 ?

Hugo-Inmanta, line 627: Slight preference for the following name (`old` implies the existence of a `..._new` method to me). And if we move this to core, the signature will probably also include the modules as argument and return value.
```suggestion
    def _unregister_modules(cls, project: module.Project):
```

Hugo-Inmanta (changes requested):
If this method moves to core I think it would be good to mention it in the `pytest_inmanta/plugin.py.ProjectLoader` class' docstring.

